PR #24, opened 2020-05-18
Update build.sh to includee a proxy option

rayzhou2017:


rayzhou2017:
@pixiake needs /approve

PR #27, opened 2020-05-19
Update cmd options

rayzhou2017:


cmd/create/config.go
```diff
@@ -20,9 +20,9 @@ func NewCmdCreateCfg() *cobra.Command {
 		},
 	}
 
-	clusterCfgCmd.Flags().StringVarP(&addons, "add", "", "", "Add plugins")
-	clusterCfgCmd.Flags().StringVarP(&name, "name", "", "", "Name of config file")
-	clusterCfgCmd.Flags().StringVarP(&clusterCfgPath, "config", "f", "", "cluster info config")
+	clusterCfgCmd.Flags().StringVarP(&addons, "add", "", "", "add plugins")
+	clusterCfgCmd.Flags().StringVarP(&name, "name", "", "", "cluster name")
+	clusterCfgCmd.Flags().StringVarP(&clusterCfgPath, "config", "f", "", "configuration file name")
```
rayzhou2017, line 25: what does the option look like? --config or -f?

cmd/create/config.go
```diff
@@ -20,9 +20,9 @@ func NewCmdCreateCfg() *cobra.Command {
 		},
 	}
 
-	clusterCfgCmd.Flags().StringVarP(&addons, "add", "", "", "Add plugins")
-	clusterCfgCmd.Flags().StringVarP(&name, "name", "", "", "Name of config file")
-	clusterCfgCmd.Flags().StringVarP(&clusterCfgPath, "config", "f", "", "cluster info config")
+	clusterCfgCmd.Flags().StringVarP(&addons, "add", "", "", "add plugins")
+	clusterCfgCmd.Flags().StringVarP(&name, "name", "", "", "cluster name")
```
rayzhou2017, line 24: is this --name? Shall we support -a?

pixiake: `-n` can be supported. But `-n` may have other meanings，such as `namespace`.  So I didn't add it.

cmd/create/config.go
pixiake: It should be modified here. The value of this parameter is the file path，So i think  `--config` should be deprecated 
or replaced with `--file` here.
I'll modify here later.

PR #29, opened 2020-05-20
Update config-example.yaml

rayzhou2017:


docs/config-example.md
```diff
@@ -22,7 +22,7 @@ spec:
     address: ""
     port: "6443"
   kubernetes:
-    version: v1.17.4
+    version: v1.17.5
     imageRepo: kubekey
```
rayzhou2017, line 26: What does imageRepo mean here? Does it mean the K8s images are from hub.docker.com/kubekey?

pixiake: Yes，I haven't updated here. `hub.docker.com/kubesphere` is default.

PR #31, opened 2020-05-20
add RBD storage support

pixiake:
Add support for RBD storage.

### DIFF
--- a/pkg/apis/kubekey/v1alpha1/default.go
+++ b/pkg/apis/kubekey/v1alpha1/default.go
@@ -41,7 +41,7 @@ func (cfg *ClusterSpec) SetDefaultClusterSpec() (*ClusterSpec, *HostGroups) {
 	clusterCfg.Network = SetDefaultNetworkCfg(cfg)
 	clusterCfg.Kubernetes = SetDefaultClusterCfg(cfg)
 	clusterCfg.Registry = cfg.Registry
-	clusterCfg.Storage = cfg.Storage
+	clusterCfg.Storage = SetDefaultStorageCfg(cfg)
 	clusterCfg.KubeSphere = cfg.KubeSphere
 	if cfg.Kubernetes.ImageRepo == "" {
 		clusterCfg.Kubernetes.ImageRepo = DefaultKubeImageRepo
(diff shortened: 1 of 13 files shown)


pkg/apis/kubekey/v1alpha1/plugins_types.go
```diff
@@ -2,17 +2,21 @@ package v1alpha1
 
 type Storage struct {
 	DefaultStorageClass string      `yaml:"defaultStorageClass" json:"defaultStorageClass,omitempty"`
-	LocalVolume         LocalVolume `yaml:"localVolume" json:"localVolume,omitempty"`
-	NfsClient           NfsClient   `yaml:"nfsClient" json:"nfsClient,omitempty"`
-	GlusterFS           GlusterFS   `yaml:"glusterFS" json:"glusterFS,omitempty"`
-	CephRBD             CephRBD     `yaml:"cephRBD" json:"cephRBD,omitempty"`
+	LocalVolume         LocalVolume `yaml:"local" json:"locaVolume,omitempty"`
+	NfsClient           NfsClient   `yaml:"nfs" json:"nfs,omitempty"`
```
rayzhou2017, line 6: why here is nfsclient, while the yaml is nfs?

pixiake: The name of this plugin is nfsclient.

pkg/apis/kubekey/v1alpha1/default.go
```diff
@@ -126,3 +126,35 @@ func SetDefaultClusterCfg(cfg *ClusterSpec) Kubernetes {
...
+		cfg.Storage.NfsClient.Enabled = true
+	}
+	if cfg.Storage.CephRBD.StorageClassName != "" {
+		cfg.Storage.CephRBD.Enabled = true
+	}
+	if cfg.Storage.GlusterFS.StorageClassName != "" {
+		cfg.Storage.GlusterFS.Enabled = true
+	}
+
+	if cfg.Storage.DefaultStorageClass != "" {
```
min-zh, line 144: if **cfg.storage.DefaultStorageClass** is none of the listed cases,  what will happen ?

pkg/apis/kubekey/v1alpha1/plugins_types.go
min-zh: [nfs](https://github.com/kubernetes-incubator/external-storage/tree/master/nfs) and [nfs-client](https://github.com/kubernetes-incubator/external-storage/tree/master/nfs-client) are two different **external-storage** of [kubernetes-incubator](https://github.com/kubernetes-incubator/external-storage) .  Actually what we use is nfs-client, i suggest use nfs-client for unity and no confoud.

pkg/apis/kubekey/v1alpha1/default.go
pixiake: No default storage classe will be set.

pkg/apis/kubekey/v1alpha1/plugins_types.go
pixiake: Yes. Then we'll still use `nfsClient`. @rayzhou2017 Does it need to be changed in readme？

rayzhou2017: yes, if it is that case, then make it consistent.

PR #137, opened 2020-07-09
Update readme

rayzhou2017:


README_zh-CN.md
```diff
@@ -75,9 +78,9 @@
     ```
 
 > 注意：
-> 
+>
 > * 在构建之前，需要先安装 Docker。
-> * 如果无法访问 `https://proxy.golang.org/`，请执行 `build.sh -p` 代替。
+> * 如果无法访问 `https://proxy.golang.org/`，比如在大陆，请执行 `build.sh -p`。
```
FeynmanZhou, line 83: `比如在大陆` 可以去掉？

rayzhou2017: not necessary, since it is Chinese I think.

rayzhou2017: If you think it is better to remove, that is fine.

PR #156, opened 2020-07-19
Update supported K8s versions

rayzhou2017:
/cc @pixiake Please also update kubekey to supported the latest versions.


pixiake:
I'll sync the new version of the kube binaries.

Issue #173, opened 2020-07-28
Make es configurable when installing ks


rayzhou2017:
Current config yaml only uses built-in es. It is highly possible for enterprise customers to configure it to user external es. So provide such configuration in the lnstallation config file.

benjaminhuo:
[config-example.md](https://github.com/kubesphere/kubekey/blob/master/docs/config-example.md)  this config-example doc is too old to reflect current changes: the es config has been moved from logging to common by [this PR](https://github.com/kubesphere/ks-installer/pull/904/commits/b7ac9240441224b575bba29425c46d58746b1a31). 

We may need to adjust cc or this doc accordingly .

There're already external es setting in ks-installer like this, we can simply add it to the cc's doc?:
![image](https://user-images.githubusercontent.com/18525465/88659441-a0255980-d107-11ea-89f0-418d4f1b4abb.png)

rayzhou2017:
verified

PR #177, opened 2020-07-30
update config example

huanggze:
Signed-off-by: huanggze <loganhuang@yunify.com>

Ref. https://github.com/kubesphere/kubekey/issues/173


huanggze:
Is metrics-server option right? @pixiake 

image

Issue #94, opened 2020-06-24
preflight check: check the iptables on the host match the iptables on the kube-proxy image


zheng1:
Related issues: 
https://github.com/kubernetes/kubernetes/issues/71305 https://github.com/kubernetes/kubernetes/issues/71305

/assign @duanjiong

PR #236, opened 2020-08-19
Make sure the iptables utility doesn't use the nftables backend.

duanjiong:
Signed-off-by: Duan Jiong <djduanjiong@gmail.com>


duanjiong:
Fix #94

PR #390, opened 2020-12-08
Use kubekey as the offical binary name instead of kk

LinuxSuRen:
Align the binary name to the previous one.

I test this via `goreleaser release --skip-publish --rm-dist`

Below is the partial log output:
```
   • archives
      • creating                  archive=bin/kubekey-v1.1.4-linux-arm64.tar.gz
      • creating                  archive=bin/kubekey-v1.1.4-linux-amd64.tar.gz
   • creating source archive
      • pipe skipped              error=source pipe is disabled
   • linux packages
      • creating                  file=bin/kubekey-v1.1.4-linux-64bit.rpm
      • creating                  file=bin/kubekey-v1.1.4-linux-arm64.deb
      • creating                  file=bin/kubekey-v1.1.4-linux-arm64.rpm
      • creating                  file=bin/kubekey-v1.1.4-linux-64bit.deb
```

**Open Question**
* Do we need Linux Packages for kk?

I find there is no Linux package for kk on [the last release page](https://github.com/kubesphere/kubekey/releases/tag/v1.1.0-alpha.1). I'm not sure we want the Linux packages: .rpm and .deb.

If we think it's unnecessary I can remove it from the config file. But in my opinion, we should keep them.

* Take a concise name?
`kubekey` is the full name of this project. I know this is an official name. But in my opinion, `kk` is much more simple and easy to remember.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,19 +27,19 @@ jobs:
         run: |
           kkver=$(hub tag --list | sort -V | tail -n 1)
           echo "start to upload rpm packages"
-          curl -T bin/kk-linux-64bit.rpm -ulinuxsuren:${{ secrets.BINTRAY_TOKEN }} \
+          curl -T bin/kubekey-${kkver}-linux-64bit.rpm -ulinuxsuren:${{ secrets.BINTRAY_TOKEN }} \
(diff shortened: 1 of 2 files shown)


pixiake:
I think the `.rpm and .deb` packages are needed.  It's just that the naming convention is different from before, so it's not shown yet.

PR #411, opened 2020-12-28
Fix the nightly installation errors

pixiake:

### DIFF
--- a/pkg/kubesphere/manifests.go
+++ b/pkg/kubesphere/manifests.go
@@ -416,11 +416,11 @@ spec:
 )
 
 func GenerateKubeSphereYaml(repo, version string) (string, error) {
-	if os.Getenv("KKZONE") == "cn" {
+	if strings.Contains(version, "latest") && os.Getenv("KKZONE") == "cn" {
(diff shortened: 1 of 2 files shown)


pkg/kubesphere/manifests.go
```diff
@@ -416,11 +416,11 @@ spec:
 )
 
 func GenerateKubeSphereYaml(repo, version string) (string, error) {
-	if os.Getenv("KKZONE") == "cn" {
+	if strings.Contains(version, "latest") && os.Getenv("KKZONE") == "cn" {
```
LinuxSuRen, line 419: How about these versions like v3.0.0, v3.1.0? Actually, I would suggest we sync all images into this registry which include the nightly build image and kubespheredev

pixiake: v3.0.0 does not support changing the namespace of images yet. Currently only the latest images are synced, this policy will be adjusted after other images are synchronized.

LinuxSuRen: Got you.

LinuxSuRen (approved):
/lgtm

PR #419, opened 2021-01-08
add init script

yuswift:
Fix network problem caused by kernel argument.
See https://imroc.io/posts/kubernetes/troubleshooting-with-kubernetes-network  & https://kubesphere.com.cn/forum/d/3383-ksaws-eks-member/17

### DIFF
--- a/pkg/cluster/preinstall/tmpl/init_os.go
+++ b/pkg/cluster/preinstall/tmpl/init_os.go
@@ -49,6 +49,9 @@ echo 'net.bridge.bridge-nf-call-ip6tables = 1' >> /etc/sysctl.conf
 echo 'net.bridge.bridge-nf-call-iptables = 1' >> /etc/sysctl.conf
 echo 'net.ipv4.ip_local_reserved_ports = 30000-32767' >> /etc/sysctl.conf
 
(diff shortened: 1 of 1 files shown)


pkg/cluster/preinstall/tmpl/init_os.go
```diff
@@ -49,6 +49,9 @@ echo 'net.bridge.bridge-nf-call-ip6tables = 1' >> /etc/sysctl.conf
 echo 'net.bridge.bridge-nf-call-iptables = 1' >> /etc/sysctl.conf
 echo 'net.ipv4.ip_local_reserved_ports = 30000-32767' >> /etc/sysctl.conf
 
+#See https://imroc.io/posts/kubernetes/troubleshooting-with-kubernetes-network/
+echo “net.ipv4.tcp_tw_recycle = 0” >> /etc/sysctl.conf
```
pixiake, line 53: I see `tcp_tw_recycle` will be removed from the kernel after version 4.12.
Will there be a problem if all kernal version are added?

pixiake:
@all-contributors please add @yuswift for code

pkg/cluster/preinstall/tmpl/init_os.go
yuswift: Good suggestion. I've updated the code.

PR #425, opened 2021-01-25
Add user defined tools download command option

LinuxSuRen:
This PR provider an extension point for downloading tools, for example, users can set the timeout, proxy, or retry under some poor network environment. Or users even can choose another cli, it might be wget. Perhaps we should have a build-in download function instead of totally rely on the external one.


LinuxSuRen:
This is an example of usage under the new feature:

`kk create cluster -y --download-cmd 'hd get -t 8 -o %s %s'`

[hd](https://github.com/LinuxSuRen/http-downloader) is a tool that can download files with multi-thread.

PR #443, opened 2021-02-02
Fix download-cmd parameter about adding nodes and updating cluster

Forest-L:
Signed-off-by: Forest-L <lilin@yunify.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.14
 
 require (
 	github.com/dominodatalab/os-release v0.0.0-20190522011736-bcdb4a3e3c2f
-	github.com/go-logr/logr v0.4.0
+	github.com/go-logr/logr v0.1.0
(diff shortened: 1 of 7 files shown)


pixiake (commented):
/lgtm

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.14
 
 require (
 	github.com/dominodatalab/os-release v0.0.0-20190522011736-bcdb4a3e3c2f
-	github.com/go-logr/logr v0.4.0
+	github.com/go-logr/logr v0.1.0
```
LinuxSuRen, line 7: It's wired. Why downgrade the version of `go-logr` here.

pixiake: operator-sdk does not support logr v0.4.0

LinuxSuRen: Got you.

PR #448, opened 2021-02-04
V1.1.0 alpha.1 branch

xww822:
接受提交


pixiake:
I think the comments should be in English.

pixiake:
kubekey provides addons to deploy custom plugins， so it is recommended to use the addons to deploy rook-ceph.
https://github.com/kubesphere/kubekey/blob/master/docs/addons.md

shaowenchen:
Thanks for your contribution. @xww822 

 However, we need to keep each pr single. so, I suggest turning off this pr and splitting it into multiples.

PR #525, opened 2021-05-04
use go std-library's SHA256Sum to check download binary file

alimy:
kk check downloaded binary file use cli tool that named sha256sum will failure in Mac OS because sha256sum is not installed default. This patch use golang std-library's `sha256.Sum256(...)` instead of sha256sum cli tool to check download binary file.


pixiake:
@all-contributors please add @alimy for code

Issue #546, opened 2021-06-02
Etcd backup failed for wrong ENDPOINTS


lshmouse:
```
# bash /usr/local/bin/kube-scripts/etcd-backup.sh
2021-06-02 09:58:55.673881 I | ignoring v3 raft entry
2021-06-02 09:58:55.673892 I | ignoring v3 raft entry
2021-06-02 09:58:55.673899 I | ignoring v3 raft entry
Error: snapshot must be requested to one selected node, not multiple [https://10.38.201.221:2379 https://10.38.201.180:2379 https://10.38.201.90:2379]

# ls -al etcd-2021-06-02_09:58:55
total 12
drwxr-xr-x 3 root root 4096 Jun  2 09:58 .
drwxr-xr-x 6 root root 4096 Jun  2 09:59 ..
drwx------ 4 root root 4096 Jun  2 09:58 member

```

lshmouse:
The dir name : etcd-2021-06-02_09:46:43 is not convenient for bash completion

lshmouse:
Also see: https://bugzilla.redhat.com/show_bug.cgi?id=1822924

LinuxSuRen:
hi @lshmouse , please help to check if #547 fixed this issue. I saw you just left the comment `Ref #546`. 

/kind bug

PR #547, opened 2021-06-02
Set Etcdendpoint to local node instead of all etcd members in etcd-backup.sh

lshmouse:
as title～

Ref https://github.com/kubesphere/kubekey/issues/546


pixiake:
Thanks for the contribution！

PR #554, opened 2021-06-04
Fix: modified the timer in run_task.go

### LINKED ISSUE #552: There may be a bug in the code that excute tasks in parallel
i'm reading the code in this file `run_task.go`.

```
if parallel {
		go func(result chan string, ccons chan struct{}) {
			for i := 0; i < hostNum; i++ {
				select {
				case <-result:
				case <-time.After(time.Minute * Timeout):
					mgr.Logger.Fatalf("Execute task timeout, Timeout=%ds", Timeout)
				}
				wg.Done()
				<-ccons
			}
		}(result, ccons)
	}
```

it will be generating a new `timer` every times in the loop. 

So when timeout takes effect, the execution result may appears:

```
// assume timeout = 120s
task 1 complete
task 2 complete
// timeout 
Execute task timeout, Timeout=120s
// generate a new timer and the timeout is be set up 120s
task 4 complete
task 5 complete
```  

I guess there should be a consistent timeout for  every task in parallel. When the timeout takes effect, All unfinished tasks should be shutdown (pherhaps there is only printing a fatal log in the code).

So, how do you think that? Maybe I can help with to modify it.

24sama:
Modified the timer in run_task.go

Resolve the problem of inconsistent task timeout control. I move the timer into the goruntine, and every goruntine has a timer. Because of that, the time of each timer will be vary close.

Which issue(s) this PR fixes:
Fix #552

### DIFF
--- a/pkg/util/manager/run_task.go
+++ b/pkg/util/manager/run_task.go
@@ -99,25 +99,8 @@ func (mgr *Manager) RunTaskOnNodes(nodes []kubekeyapiv1alpha1.HostCfg, task Node
 	hasErrors := false
 
 	wg := &sync.WaitGroup{}
-	result := make(chan string)
 	ccons := make(chan struct{}, DefaultCon)
-	defer close(result)
 	defer close(ccons)
-	hostNum := len(nodes)
-
-	if parallel {
-		go func(result chan string, ccons chan struct{}) {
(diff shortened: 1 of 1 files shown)


pixiake (commented):
/lgtm

pixiake:
@all-contributors please add @24sama for code

PR #563, opened 2021-06-21
Add kube-vip as the internal load balancer

24sama:
Add kube-vip as the internal load balancer

I add 2 tasks of creating cluster that run on the master node. The task one `InstallLoadBalancer` generate a manifests yaml file, and save it in the `/etc/kubernetes/manifests/`. Then the kubelet will start up it as a static pod. Another task is uesd to ensure the manifests file is exists on every master node.


pkg/kubernetes/preinstall/prepull_images.go
```diff
@@ -73,6 +74,8 @@ func GetImage(mgr *manager.Manager, name string) images.Image {
 		// storage
 		"provisioner-localpv": {RepoAddr: mgr.Cluster.Registry.PrivateRegistry, Namespace: "openebs", Repo: "provisioner-localpv", Tag: "2.9.0", Group: kubekeyapiv1alpha1.Worker, Enable: false},
 		"linux-utils":         {RepoAddr: mgr.Cluster.Registry.PrivateRegistry, Namespace: "openebs", Repo: "linux-utils", Tag: "2.9.0", Group: kubekeyapiv1alpha1.Worker, Enable: false},
+		// load balancer
+		"kubevip": {RepoAddr: mgr.Cluster.Registry.PrivateRegistry, Namespace: "plndr", Repo: "kube-vip", Tag: "v0.3.5", Group: kubekeyapiv1alpha1.Master, Enable: true},
```
pixiake, line 78: I think this image is only pulled when `internalLoadBalancerEnabled` is `true`.

pkg/cluster/install/install.go
```diff
@@ -98,8 +98,10 @@ func ExecTasks(mgr *manager.Manager) error {
 		{Task: etcd.BackupEtcd, ErrMsg: "Failed to backup etcd data"},
 		{Task: GetClusterStatus, ErrMsg: "Failed to get cluster status"},
 		{Task: InstallKubeBinaries, ErrMsg: "Failed to install kube binaries"},
+		{Task: InstallLoadBalancer, ErrMsg: "Failed to install load balancer", Skip: isK3s || len(mgr.MasterNodes) == 1},
```
pixiake, line 101: K3s should also be supported.

24sama: oh, i will do it.

PR #564, opened 2021-06-28
fix: modify the shell scripts

24sama:
Use the `command -v tar` to check if the `tar` command is existed. And then, echo the diffrent  prompt.


24sama (commented):
done

PR #567, opened 2021-06-30
Feature: Add internal LB haproxy for kubernetes.

24sama:
This PR is to add a HA plan for creating cluster. This plan is the same as this repo [kubespray](https://github.com/kubernetes-sigs/kubespray/blob/master/docs/ha-mode.md). 

In short, the kubelet of each master nodes connects the local apiserver and the kubelet of each worker nodes connects the local haproxy server that proxy the traffic to every apiserver.

Now it's only support kubernetes, i will support the k3s as soon as possible.

### DIFF
--- a/pkg/kubernetes/config/v1beta2/kubeadm.go
+++ b/pkg/kubernetes/config/v1beta2/kubeadm.go
@@ -63,6 +63,9 @@ networking:
   serviceSubnet: {{ .ServiceSubnet }}
 apiServer:
   extraArgs:
+    {{- if .InternalLBDisabled }}
+    advertise-address: {{ .AdvertiseAddress }}
+    {{- end }}
 {{ toYaml .ApiServerArgs | indent 4}}
   certSANs:
     {{- range .CertSANs }}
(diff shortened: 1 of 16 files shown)


pkg/kubernetes/config/v1beta2/kubeadm.go
```diff
@@ -63,7 +63,9 @@ networking:
   serviceSubnet: {{ .ServiceSubnet }}
 apiServer:
   extraArgs:
+    {{- if .InternalLBDisabled }}
```
pixiake, line 66: `advertise-address` shouldn't exist here，this will result in only one api-endpoint.

pkg/kubernetes/nodes.go
```diff
@@ -131,3 +132,93 @@ func SetKubelet(mgr *manager.Manager, node *kubekeyapiv1alpha1.HostCfg) error {
...
+			if err != nil {
+				return errors.Wrap(errors.WithStack(err), "Failed to find /etc/kubernetes/admin.conf")
+			}
+			return errors.New("Failed to find /etc/kubernetes/admin.conf")
+		}
+	}
+
+	output2, err := mgr.Runner.ExecuteCmd("sudo -E /bin/sh -c \"[ -f ~/.kube/config ] && echo 'kubectl config is exists.' || echo 'kubectl config is not exists.'\"", 0, false)
+	if strings.Contains(output2, "kubectl config is exists.") {
+		if _, err := mgr.Runner.ExecuteCmd(fmt.Sprintf("sudo sed -i 's#server:.*#server: https://127.0.0.1:%s#g' ~/.kube/config", strconv.Itoa(mgr.Cluster.ControlPlaneEndpoint.Port)), 0, false); err != nil {
```
pixiake, line 214: I think it is possible to reduce the `sed` operation by modifying the rule (`127.0.0.1  lb.kubesphere.local`) in `/etc/hosts`.

Issue #602, opened 2021-08-02
flannel vxlan+Directrouting not working


shake:
HI 

I try to test flannel vxlan + Directrouting, 

  network:
    plugin: flannel
    flannel:
        backendMode: vxlan
        Directrouting: true

after deploy, not working. only vxlan.

 net-conf.json: |
    {
      "Network": "10.233.64.0/18",
      "Backend": {
        "Type": "vxlan"
      }
    }
---

rayzhou2017:
@zheng1 Please also check this issue.

pixiake:
https://github.com/kubesphere/kubekey/blob/fb0ad3ab9c613e8be318bd521c1d0ef2f48e1028/apis/kubekey/v1alpha1/network_types.go#L34

For flannel, at present, it only supports specifying `Backend`, but does not support the configuration of `Directrouting`.

shake:
ok, good idea. let my team try it.  this is Very good opportunity to enhance our ability.

PR #606, opened 2021-08-04
Add support for flannel vxlan+Directrouting

kinglong08:


kinglong08 (commented):
Add support flannel vxlan+Directrouting

FeynmanZhou:
I thinks this PR is fixing https://github.com/kubesphere/kubekey/issues/602

shake:
for flannel , vxlan support Directrouting, ipip also Directrouting.

https://github.com/flannel-io/flannel/blob/master/Documentation/backends.md

apis/kubekey/v1alpha1/network_types.go
```diff
@@ -32,7 +32,12 @@ type CalicoCfg struct {
 }
 
 type FlannelCfg struct {
-	BackendMode string `yaml:"backendMode" json:"backendMode,omitempty"`
+	BackendMode string     `yaml:"backendMode" json:"backendMode,omitempty"`
+	Backend     BackendCfg `yaml:"backend" json:"backend,omitempty"`
+}
+type BackendCfg struct {
```
24sama, line 38: According to this struct, we need to write our config yaml file like this:
```
network:
  plugin: flannel
  kubePodsCIDR: 10.233.64.0/18
  flannel:
    backendMode: vxlan
    backend:
      directRouting: true
```
So, I think it might be better just like the @shake issue comment #606:
```
network:
  plugin: flannel
  flannel:
    backendMode: vxlan
    directrouting: true
```

kinglong08: I tested it successfully like this.

```
network:
  plugin: flannel
  kubePodsCIDR: 10.233.64.0/18
  flannel:
    backend:
      type: vxlan
      directRouting: true
```

24sama: Your code is definitely correct.

24sama: The result of kk running is definitely correct. 

But according to your config yaml file will assign `vxlan` to the `type` field and the `BackendMode` is an empty string. 

In the `pkg/plugins/network/flannel/tmpl.go` the variable `{{ .BackendMode }}` value is `vxlan` caused by there is a default const named by `DefaultBackendMode` assigned to `BackendMode` when kk init.

In terms of convenience, i think it will more simple to confiure by the second way.

FeynmanZhou:
How's this PR going?

kinglong08 (commented):
Update support for flannel vxlan+Directrouting.
yaml config like this:

network:
    plugin: flannel
    flannel:
      backendMode: vxlan
      directRouting: true

24sama (commented):
Thanks for your contribution :wave::wave::wave:

PR #657, opened 2021-09-07
Supports installing docker and conatinerd using binaries

pixiake:
1. Supports automatic installation of docker / containerd /crictl according to the configuration of containerManager.
2. Automatically set cgroupDriver of kubelet according to cgroupDriver of each node‘s container runtime.
3. Remove global variables in installation, scaling, upgrade and renew crets.
4. Fix docker version parsing problem during upgrade.
5. Default kubernetes version: v1.20.10


pkg/util/manager/manager.go
```diff
@@ -22,40 +22,83 @@ import (
...
+	BootstrapToken string
+	CertificateKey string
+}
+
+// UpgradeStatus is used to store cluster upgrade status
+type UpgradeStatus struct {
+	CurrentVersions   map[string]string
+	CurrentVersionStr string
+	NextVersionStr    string
+	MU                sync.Mutex
```
24sama, line 85: The mgr will deep copy itself when exec task parallel, so the mutex should be define as a pointer. Otherwise, i think it will generate a new mutex after mgr.Copy().

pixiake: Done

FeynmanZhou:
> 1. Supports automatic installation of docker / containerd /crictl according to the configuration of containerManager.
> 2. Automatically set cgroupDriver of kubelet according to cgroupDriver of each node‘s container runtime.
> 3. Remove global variables in installation, scaling, upgrade and renew crets.
> 4. Fix docker version parsing problem during upgrade.
> 5. Default kubernetes version: v1.20.10
> 
> Signed-off-by: pixiake [guofeng@yunify.com](mailto:guofeng@yunify.com)

Is it suitable to include multiple updates in one PR? IMO, it is not efficient for reviewers.

FeynmanZhou:
@kubesphere/sig-installation  Do we need to release an Alpha version for this PR?

pixiake:
> Do we need to release an Alpha version for this PR?

Yes, I think we can release the alpha version with KubeSphere.

Issue #660, opened 2021-09-09
kubekey create cluster failed when no default router in system


liulangwa:
### What is version of KubeKey has the issue?

v1.1.1

### What is your os environment?

CentOS Linux release 7.6.1810 (Core)

### A clear and concise description of what happend.

create cluster failed when no default route in system
安装节点没有默认路由，导致安装失败

### Relevant log output

```shell
RRO[17:01:11 CST] Failed to upload kubeadm certs: Failed to exec command: sudo -E /bin/sh -c "/usr/local/bin/kubeadm init phase upload-certs --upload-certs" 
W0903 17:01:06.634674  106039 common.go:148] WARNING: could not obtain a bind address for the API Server: no default routes found in "/proc/net/route" or "/proc/net/ipv6_route"; using: 0.0.0.0
W0903 17:01:06.734958  106039 version.go:102] could not fetch a Kubernetes version from the internet: unable to get URL "https://dl.k8s.io/release/stable-1.txt": Get "https://dl.k8s.io/release/stable-1.txt": dial tcp: lookup dl.k8s.io on 114.114.114.114:53: dial udp 114.114.114.114:53: connect: network is unreachable
W0903 17:01:06.734986  106039 version.go:103] falling back to the local client version: v1.20.6
cannot use "0.0.0.0" as the bind address for the API Server
To see the stack trace of this error execute with --v=5 or higher: Process exited with status 1  node=172.28.8.204
WARN[17:01:11 CST] Task failed ...                              
WARN[17:01:11 CST] error: interrupted by error                  
Error: Failed to init kubernetes cluster: interrupted by error


----

[root@compute-1 kubesphere]# /usr/local/bin/kubeadm init phase upload-certs  --upload-certs --v=5
I0903 17:06:52.506505  108080 initconfiguration.go:104] detected and using CRI socket: /var/run/dockershim.sock
W0903 17:06:52.506721  108080 common.go:148] WARNING: could not obtain a bind address for the API Server: no default routes found in "/proc/net/route" or "/proc/net/ipv6_route"; using: 0.0.0.0
I0903 17:06:52.610073  108080 version.go:185] fetching Kubernetes version from URL: https://dl.k8s.io/release/stable-1.txt
[...]

PR #661, opened 2021-09-09
fix:create cluster failed when no default router in system

liulangwa:


liulangwa:
#660

pixiake:
Thanks for your contribution.

PR #695, opened 2021-09-24
dockerfile build from golang:1.16

24sama:
Change the dockerfile image and build.sh docker image to golang:1.16.

Github actions checkout add config just like kubesphere:
```
with:
  fetch-depth: 0
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,8 @@ jobs:
 
       - name: Check out code into the Go module directory
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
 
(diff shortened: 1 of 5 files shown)


pixiake:
https://github.com/kubesphere/kubekey/blob/1ddddba1b3784a5548770474d536df3bc79cd13b/Makefile#L102

The Go version should also be upgraded in Makefile.

24sama:
> https://github.com/kubesphere/kubekey/blob/1ddddba1b3784a5548770474d536df3bc79cd13b/Makefile#L102
> 
> The Go version should also be upgraded in Makefile.

PR #765, opened 2021-11-14
dev-v2.0.0: refactor cmd package

24sama:
### What dose this PR do?
* Refactor cmd package

### Why we need this PR?
KubeKey need to support plugin mode, and the old cmd package is not flexible. So i refactor KubeKey cmd package according to the design of  `kubectl` cmd package.


pixiake:
Wonderful ~ 

/lgtm

PR #788, opened 2021-11-22
dev-v2.0.0: fix a kube config file owner bug

24sama:
### What does this PR do?
* fix a kube config file ownner bug
* fix a tar file that extracts to a wrong dir bug

### DIFF
--- a/pkg/certs/tasks.go
+++ b/pkg/certs/tasks.go
@@ -328,18 +328,30 @@ type SyneKubeConfigToWorker struct {
 }
 
 func (s *SyneKubeConfigToWorker) Execute(runtime connector.Runtime) error {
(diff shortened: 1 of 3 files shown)


pixiake:
I think it needs to be modified here, too.

https://github.com/kubesphere/kubekey/blob/f960dcff0bdaaf8c624afc8f2634016dd165951d/pkg/k3s/tasks.go#L116

24sama:
> I think it needs to be modified here, too.
> 
> https://github.com/kubesphere/kubekey/blob/f960dcff0bdaaf8c624afc8f2634016dd165951d/pkg/k3s/tasks.go#L116

Ok, done.

PR #865, opened 2021-12-10
fix(plugins): cilium should deloyed with new k8s(1.16+)

tanguofu:
there is bug: cilium is skiped when use k8s 1.21.5;

### DIFF
--- a/pkg/plugins/network/modules.go
+++ b/pkg/plugins/network/modules.go
@@ -176,7 +176,7 @@ func deployCilium(d *DeployNetworkPluginModule) []task.Interface {
 		Hosts: d.Runtime.GetHostsByRole(common.Master),
 		Prepare: &prepare.PrepareCollection{
 			new(common.OnlyFirstMaster),
-			new(OldK8sVersion),
+			&OldK8sVersion{Not: true},
 		},
 		Action: &action.Template{
 			Template: templates.Cilium,


24sama:
Thanks for your contribution!

pkg/plugins/network/modules.go
```diff
@@ -176,7 +176,7 @@ func deployCilium(d *DeployNetworkPluginModule) []task.Interface {
 		Hosts: d.Runtime.GetHostsByRole(common.Master),
 		Prepare: &prepare.PrepareCollection{
 			new(common.OnlyFirstMaster),
-			new(OldK8sVersion),
+			&OldK8sVersion{Not: true},
```
24sama, line 179: I think `cilium` and `flannel` can remove this `prepare` named `OldK8sVersion`.

https://github.com/kubesphere/kubekey/blob/255c21afef3a25c132d54b91f9bebe271e0b80f5/pkg/plugins/network/modules.go#L142

24sama (changes requested):

pkg/plugins/network/modules.go
tanguofu: so do i need recommit a new PR with delete the `OldK8sVersion`  within `prepare` ?

24sama: You can push a new commit to modify it and then rebase these commits.

tanguofu:
re PR with #866

PR #882, opened 2021-12-14
Fix savekubeconfig： use InternalAddress that can't be connected when kk is not running in k8s node internal network.

tanguofu:
tested with tentcent cvm


24sama:
Thanks for your contribution!

PR #898, opened 2021-12-17
use raw string for regexp to avoid escape twice.

haiker2011:
Signed-off-by: sunhaizhou [孙海洲] <sunhaizhou@kingsoft.cn>


24sama:
Thanks for your contribution.

/lgtm
/approve

Issue #913, opened 2021-12-21
proposal :   disable SELINUX in init the OS script


tanguofu:
### Your current KubeKey version

2.0.0

### Describe this feature

it is recommend that [disable SELINUX when use kubeadmin deploy a k8s cluster](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl)

  

> Setting SELinux in permissive mode by running setenforce 0 and sed ... effectively disables it. This is required to allow containers to access the host filesystem, which is needed by pod networks for example. You have to do this until SELinux support is improved in the kubelet.

> You can leave SELinux enabled if you know how to configure it but it may require settings that are not supported by kubeadm.

### Describe the solution you'd like

this propose add the disable in init os
```
if [ -f /etc/selinux/config ]; then 
  sed -ri 's/SELINUX=enforcing/SELINUX=disabled/' /etc/selinux/config
fi
# for ubuntu: sudo apt install selinux-utils
# for centos: yum install selinux-policy
if command -v setenforce &> /dev/null
then
  setenforce 0
  getenforce
fi
```

24sama:
fixes: #914

PR #914, opened 2021-12-21
fea(#913): [kk] disable SELINUX in init the OS

tanguofu:


pixiake:
Thanks for this contribution

/approve
/lgtm

Issue #915, opened 2021-12-22
proposal ：support  add  private registry authorization when kk deloy k8s


tanguofu:
### Your current KubeKey version

2.0.0

### Describe this feature

In a private cloud environment，The docker registry  always is private ， need authorization  to login，this feature add `auths` config at `registry` section of [config-example.md]（https://github.com/kubesphere/kubekey/blob/master/docs/config-example.md）
```
...
registry:
  auths:
   "registry-1.docker.io":
      username: xxx
      password: ***
...
```

### For docker:
As kubelet will read the $HOME/.docker/json , see [Configuring nodes to authenticate to a private registry](https://kubernetes.io/docs/concepts/containers/images/#configuring-nodes-to-authenticate-to-a-private-registry), just run 
```
docker login --username xxx  --password *** registry-1.docker.io
```
with the root user

### For containerd
Add `auths`  into  `/etc/containerd/config.toml `, see [Configure Image Registry](https://github.com/containerd/cri/blob/master/docs/registry.md#configure-image-registry)
```
...
[plugins]
  [plugins."io.containerd.grpc.v1.cri"]
    sandbox_image = "kubesphere/pause:3.4.1"
    ....
    [plugins."io.containerd.grpc.v1.cri".registry]
      ...
      [plugins."io.containerd.grpc.v1.cri".registry.configs."registry-1.docker.io".auth]  
        username = "xxx"
        password = "***"
```

This action will take place at the last task in  `InstallDocker` or  `InstallContainerd` of `InstallContainerModule`

pixiake:
`/var/lib/kubelet/config.json`, Can't we use this method uniformly ?

tanguofu:
> `/var/lib/kubelet/config.json`, Can't we use this method uniformly ?

the path is RootDirectory of kubelet， it's only canbe change when set kubelet args, and here will add check , if user change kubelet 's RootDirectory , the `/var/lib/kubelet/config.json` will be change as same as user config.

tanguofu:
> `/var/lib/kubelet/config.json`, Can't we use this method uniformly ?

as  [Configuring nodes to authenticate to a private registry](https://kubernetes.io/docs/concepts/containers/images/#configuring-nodes-to-authenticate-to-a-private-registry), now will copy to  `/.docker/config.json`

PR #921, opened 2021-12-23
fea(#915): support custom private registry authorization

tanguofu:
1. for docker: fisrt docker login, then cp ~/.docker/config.json /.docker/config.json
2. for containerd: append auths into /etc/containerd/config.toml


24sama:
Thanks for your contribution!

PR #923, opened 2021-12-23
Support kubekey to independently generate certificates

pixiake:
### What does this PR do ？
* Generate certificates logic using code refactoring.
* This method can be used uniformly to generate certificates when installing other software requiring certificates by KubeKey.


24sama:
This PR will make kk more simple to manage certs.

PR #944, opened 2021-12-29
Support kubekey to independently generate certificates

pixiake:
### What does this PR do ？
* Generate certificates logic using code refactoring.
* This method can be used uniformly to generate certificates when installing other software requiring certificates by KubeKey.


pkg/utils/certs/utils.go
```diff
@@ -0,0 +1,320 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
24sama, line 2: Does the copyright need to be modified?

pixiake: The code in this file is obtained from kubernetes. So the copyright needs to be consistent with kubernetes.

pkg/etcd/certs.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+type FetchCerts struct {
+	common.KubeAction
+}
+
+func (f *FetchCerts) Execute(runtime connector.Runtime) error {
+	src := "/etc/ssl/etcd/ssl"
+	dst := fmt.Sprintf("%s/pki/etcd", runtime.GetWorkDir())
+	if v, ok := f.PipelineCache.Get(common.ETCDCluster); ok {
```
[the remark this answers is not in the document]
pixiake: Done

pkg/utils/certs/utils.go
24sama: So, is it normal for a project to have two different copyright?

PR #947, opened 2021-12-30
fix cmd delete node.

haiker2011:
fix cmd delete `nodes` to `node`, now only support delete one node.


pixiake:
Thanks for this contribution.

/lgtm
/approve

PR #952, opened 2022-01-01
fea(#915): support set ntp server and timezone when kk deloy k8s

tanguofu:


pkg/bootstrap/os/tasks.go
```diff
@@ -454,3 +454,32 @@ func (u *UmountISO) Execute(runtime connector.Runtime) error {
...
 	return nil
 }
+
+type NodeConfigureNtpServer struct {
+	common.KubeAction
+}
+
+func (n *NodeConfigureNtpServer) Execute(runtime connector.Runtime) error {
+
+	for _, server := range n.KubeConf.Cluster.System.NtpServers {
```
pixiake, line 464: I think if no set time server, we can selecte one  node in the cluster as the default time server, and other nodes can synchronize time from this node. What about you ?

tanguofu: ok i think  we can  select master node as ntp servers to avoid single point of failure.

tanguofu: @pixiake  now support use k8s node to configure the ntp server, please review again.

24sama: What happened when the length of ntpServer is 0?

tanguofu: > What happened when the length of NtpServers is 0? Do nothing, just like skip this task? Maybe you can add a `prepare` to this task. That will make the log print more graceful, because they will show `[node1]:skipped` when the condition is false.

got, i had fix this with add `NodeConfigureNtpCheck` to check the config, please review again @24sama

pkg/bootstrap/os/tasks.go
```diff
@@ -454,3 +454,71 @@ func (u *UmountISO) Execute(runtime connector.Runtime) error {
...
 	return nil
 }
+
+type NodeConfigureNtpCheck struct {
+	common.KubePrepare
+}
+
+func (n *NodeConfigureNtpCheck) PreCheck(runtime connector.Runtime) (bool, error) {
+
+	if len(n.KubeConf.Cluster.System.NtpServers) == 0 && len(n.KubeConf.Cluster.System.Timezone) == 0 {
```
24sama, line 464: if the user didn't set the `timezone`, this task will skip, does it equal your expectation?

pkg/bootstrap/os/tasks.go
```diff
@@ -454,3 +454,71 @@ func (u *UmountISO) Execute(runtime connector.Runtime) error {
...
+		if _, err := runtime.GetRunner().SudoCmd(setTimeZoneCmd, false); err != nil {
+			return errors.Wrapf(err, "set timezone: %s failed", n.KubeConf.Cluster.System.Timezone)
+		}
+
+		if _, err := runtime.GetRunner().SudoCmd("timedatectl set-ntp true", false); err != nil {
+			return errors.Wrap(err, "timedatectl set-ntp true failed")
+		}
+	}
+
+	if len(n.KubeConf.Cluster.System.NtpServers) > 0 || len(n.KubeConf.Cluster.System.Timezone) > 0 {
```
24sama, line 514: There is a `prepare` before this `action`, I think `len(n.KubeConf.Cluster.System.NtpServers) > 0` it can be removed

pkg/bootstrap/os/tasks.go
```diff
@@ -454,3 +454,71 @@ func (u *UmountISO) Execute(runtime connector.Runtime) error {
 	}
 	return nil
 }
+
+type NodeConfigureNtpCheck struct {
```
24sama, line 458: You can add a `prepares.go` file in this dir, and put the `prepare` code in it.

pkg/bootstrap/os/tasks.go
24sama: Ok, Thanks again. @tanguofu

pkg/bootstrap/os/tasks.go
tanguofu: if the user didnot set timezone but set NtpServers, the task also will be executed.

PR #953, opened 2022-01-01
fix: fix --debug not print the DEBUG LEVEL LOG

tanguofu:


24sama:
Thanks for your contribution!

PR #955, opened 2022-01-03
Creating container images registry is supported

pixiake:
### What does this PR do ？
Support installation of container images registry (`docker registry` or `harbor`) using kubekey.

### How to use ?
Initialize a local images registry.
```
./kk init registry [(-f | --filename) path]
```
Add `registry` to `spec.roleGroups` of the configuration file.
example:
```
kind: Cluster
metadata:
  name: sample
spec:
  hosts:
  - {name: node1, address: 192.168.6.6, internalAddress: 192.168.6.6, password: Qcloud@123}
  roleGroups:
    etcd:
    - node1
    master:
    - node1
    worker:
    - node1
    ## Specify the node role as registry. Only one node can be set as registry.
    registry:
    - node1
  controlPlaneEndpoint:
    ##Internal loadbalancer for apiservers
    #internalLoadbalancer: haproxy

    domain: lb.kubesphere.local
    address: ""
    port: 6443
  kubernetes: 
    version: v1.21.5
    clusterName: cluster.local
  network:
    plugin: calico
    kubePodsCIDR: 10.233.64.0/18
    # multus support. https://github.com/k8snetworkplumbingwg/multus-cni
    enableMultusCNI: false
  registry:
    ## `docker registry` is used to create local registry by default.  
    # type: ""  
    registryMirrors: []
    insecureRegistries: []
  addons: []
```


cmd/ctl/init/init_registry.go
```diff
@@ -0,0 +1,67 @@
+/*
+Copyright 2020 The KubeSphere Authors.
```
haiker2011, line 2: Should 2021 or 2022 ?

pixiake: Done. Has been changed to 2022.

pkg/binaries/tasks.go
```diff
@@ -135,3 +135,21 @@ func (a *ArtifactDownload) Execute(runtime connector.Runtime) error {
...
 }
+
+type RegistryPackageDownload struct {
+	common.KubeAction
+}
+
+func (k *RegistryPackageDownload) Execute(runtime connector.Runtime) error {
+	arch := runtime.GetHostsByRole(common.Registry)[0].GetArch()
+
+	packageDir := filepath.Join(runtime.GetWorkDir(), arch)
```
24sama, line 146: How do you think modifying the path as `kubekey/registry/amd64/xxxx.tar.gz`? That will be the same as the kk k8s binaries path(`kubekey/v1.21.5/amd64/xxxx.tar.gz`).

pixiake: It is not recommended to put it in the k8s version directory. If there are multiple versions in the offline package, the files will be saved repeatedly.

pixiake: I think these files (helm / etcd / cni / docker / crictl ...) should also be stored in a separate directory.

24sama: > It is not recommended to put it in the k8s version directory. If there are multiple versions in the offline package, the files will be saved repeatedly.

I mean put it in a dir which named `registry` not in the k8s version dir.

pixiake: Done

PR #992, opened 2022-01-11
add pod pid limit and pid available

vincenthe11:
### What does this PR do?
add pid limit and pid available for pod.

Kubernetes allows you to limit the number of processes running in a Pod. You specify this limit at the node level, rather than configuring it as a resource limit for a particular Pod. Each Node can have a different PID limit.

To configure the limit, you can specify the command line parameter set PodPidsLimit in the kubelet configuration file.
https://kubernetes.io/docs/concepts/policy/pid-limiting/#pod-pid-limits
https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#eviction-signals

### DIFF
--- a/pkg/kubernetes/templates/v1beta2/kubeadm_config.go
+++ b/pkg/kubernetes/templates/v1beta2/kubeadm_config.go
@@ -187,8 +187,10 @@ func GetKubeletConfiguration(runtime connector.Runtime, kubeConf *common.KubeCon
 			"cpu":    "200m",
 			"memory": "250Mi",
 		},
+		"podPidsLimit": 1000,
 		"evictionHard": map[string]string{
 			"memory.available": "5%",
+			"pid.available": "10%",
 		},
 		"evictionSoft": map[string]string{
 			"memory.available": "10%",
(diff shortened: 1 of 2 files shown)


LinuxSuRen (commented):
hi @vincenthe11 , thanks for your contribution. It would be great if you can provide more information on this PR.

vincenthe11:
Kubernetes allows you to limit the number of processes running in a Pod. You specify this limit at the node level, rather than configuring it as a resource limit for a particular Pod. Each Node can have a different PID limit.
To configure the limit, you can specify the command line parameter set PodPidsLimit in the kubelet configuration file.
https://kubernetes.io/docs/concepts/policy/pid-limiting/#pod-pid-limits
https://kubernetes.io/docs/concepts/scheduling-eviction/node-pressure-eviction/#eviction-signals

pkg/kubernetes/templates/v1beta2/kubeadm_config.go
```diff
@@ -187,8 +187,10 @@ func GetKubeletConfiguration(runtime connector.Runtime, kubeConf *common.KubeCon
 			"cpu":    "200m",
 			"memory": "250Mi",
 		},
+		"podPidsLimit": 1000,
 		"evictionHard": map[string]string{
 			"memory.available": "5%",
+			"pid.available": "10%",
```
LinuxSuRen, line 193: Are there any potential issues if keep using the default value?

pixiake: In some cluster, because the pod takes up a large amount of pid, the pid of nodes is exhausted, which leads to abnormal cluster. This kind of resource limiting helps to prevent simple fork bombs from affecting operation of an entire cluster. 
These parameters are @vincenthe11 accumulated experience in clusters operation.

pixiake:
@all-contributors please add @vincenthe11 for code

PR #1003, opened 2022-01-13
Support custom dnsDomain for cluster

pixiake:
### What does this PR do?
Support custom `dnsDomain` for cluster.


RolandMa1986:
@pixiake Do we need to change this function too?
https://github.com/kubesphere/kubekey/blob/88c26e54a1eb6b465c580c7f939480b62c021ea1/apis/kubekey/v1alpha2/cluster_types.go#L197

pixiake:
> @pixiake Do we need to change this function too?
> 
> https://github.com/kubesphere/kubekey/blob/88c26e54a1eb6b465c580c7f939480b62c021ea1/apis/kubekey/v1alpha2/cluster_types.go#L197


Yes, it needs to be modified. I have changed it.

PR #1010, opened 2022-01-15
Support Kata and Node Feature Discovery

pixiake:
### What does this PR do ？
Support installation of [kata-container](https://github.com/kata-containers/kata-containers) and [node-feature-discovery](https://github.com/kubernetes-sigs/node-feature-discovery) using kubekey.

Kata Containers is an open source community working to build a secure container runtime with lightweight virtual machines that feel and perform like containers, but provide stronger workload isolation using hardware virtualization technology as a second layer of defense.

NodeFeatureDiscovery detects hardware features available on each node in a Kubernetes cluster, and advertises those features using node labels.

### How to use ?
add `spec.kubernetes.enableKata: true` or  `spec.kubernetes.enableNodeFeatureDiscovery: true`
* Kata will only take effect if the container manager uses `containerd` and `cri-o`.

example:
```
kind: Cluster
metadata:
  name: sample
spec:
  hosts:
  - {name: node1, address: 192.168.6.6, internalAddress: 192.168.6.6, password: Qcloud@123}
  roleGroups:
    etcd:
    - node1
    master:
    - node1
    worker:
    - node1
    ## Specify the node role as registry. Only one node can be set as registry.
    registry:
    - node1
  controlPlaneEndpoint:
    ##Internal loadbalancer for apiservers
    #internalLoadbalancer: haproxy

    domain: lb.kubesphere.local
    address: ""
    port: 6443
  kubernetes: 
    version: v1.21.5
    clusterName: cluster.local
    enableKata: true      # enable kata-containers，and specify `runtimeClassName: kata-qemu` in the Pod spec to create kata sandbox.
    enableNodeFeatureDiscovery: true  # enable node-feature-discovery
  network:
    plugin: calico
    kubePodsCIDR: 10.233.64.0/18
    # multus support. https://github.com/k8snetworkplumbingwg/multus-cni
    enableMultusCNI: false
  registry:
    ## `docker registry` is used to create local registry by default.  
    # type: ""  
    registryMirrors: []
    insecureRegistries: []
  addons: []
```


pkg/images/tasks.go
```diff
@@ -118,9 +118,12 @@ func GetImage(runtime connector.ModuleRuntime, kubeConf *common.KubeConf, name s
 		// storage
 		"provisioner-localpv": {RepoAddr: kubeConf.Cluster.Registry.PrivateRegistry, Namespace: "openebs", Repo: "provisioner-localpv", Tag: "2.10.1", Group: kubekeyv1alpha2.Worker, Enable: false},
 		"linux-utils":         {RepoAddr: kubeConf.Cluster.Registry.PrivateRegistry, Namespace: "openebs", Repo: "linux-utils", Tag: "2.10.0", Group: kubekeyv1alpha2.Worker, Enable: false},
-
 		// load balancer
 		"haproxy": {RepoAddr: kubeConf.Cluster.Registry.PrivateRegistry, Namespace: "library", Repo: "haproxy", Tag: "2.3", Group: kubekeyv1alpha2.Worker, Enable: kubeConf.Cluster.ControlPlaneEndpoint.IsInternalLBEnabled()},
+		// kata-deploy
+		"kata-deploy": {RepoAddr: kubeConf.Cluster.Registry.PrivateRegistry, Namespace: kubekeyv1alpha2.DefaultKubeImageNamespace, Repo: "kata-deploy", Tag: "stable", Group: kubekeyv1alpha2.Worker, Enable: kubeConf.Cluster.Kubernetes.EnableKata},
```
24sama, line 124: Are there only worker nodes that can use the `kata containers`?

apis/kubekey/v1alpha2/kubernetes_types.go
```diff
@@ -29,21 +29,23 @@ type Kubernetes struct {
...
-	KubeletConfiguration     runtime.RawExtension `yaml:"kubeletConfiguration" json:"kubeletConfiguration,omitempty"`
-	KubeProxyConfiguration   runtime.RawExtension `yaml:"kubeProxyConfiguration" json:"kubeProxyConfiguration,omitempty"`
+	Nodelocaldns               *bool                `yaml:"nodelocaldns" json:"nodelocaldns,omitempty"`
+	EtcdBackupDir              string               `yaml:"etcdBackupDir" json:"etcdBackupDir,omitempty"`
+	EtcdBackupPeriod           int                  `yaml:"etcdBackupPeriod" json:"etcdBackupPeriod,omitempty"`
+	KeepBackupNumber           int                  `yaml:"keepBackupNumber" json:"keepBackupNumber,omitempty"`
+	EtcdBackupScriptDir        string               `yaml:"etcdBackupScript" json:"etcdBackupScript,omitempty"`
+	ContainerManager           string               `yaml:"containerManager" json:"containerManager,omitempty"`
+	ContainerRuntimeEndpoint   string               `yaml:"containerRuntimeEndpoint" json:"containerRuntimeEndpoint,omitempty"`
+	EnableNodeFeatureDiscovery bool                 `yaml:"enableNodeFeatureDiscovery" json:"enableNodeFeatureDiscovery,omitempty"`
```
24sama, line 39: How do you think to use a `struct` to configure the `kata` and `NodeFeatureDiscovery`?
Just like:
```
Kata:
  enable: true
  moreOptions: xxxx
NodeFeatureDiscovery:
  enable: false
```
BTW, if we use this structure, in the golang struct, the type of the `enable` field we can set as `*bool`. This allows you to distinguish between "true", "false" and "not set".

pixiake: Good idea, I will update it.

pkg/images/tasks.go
pixiake: Yes, kata does not support host network, the control-plane components uses runc, so kata is not needed on the  control-plane.

PR #1048, opened 2022-02-09
fix a bug that "--with-local-storage" flag didn't work

24sama:
### Whta does this PR do?
* fix a bug that "--with-local-storage" flag didn't work
* generate the deepcopy code

In the past, if i wanted to use the `addons` to deploy a storageclass and skip this module, i tried to use command with flag `--with-local-storage=false` but it didn't work. Because this condition didn't have a correct logic:
```
&storage.DeployLocalVolumeModule{Skip: !runtime.Arg.DeployLocalStorage && !runtime.Cluster.KubeSphere.Enabled},
```


pixiake (commented):
/lgtm

PR #1101, opened 2022-03-01
add manifest and artifact en doc

24sama:
### What type of PR is this?

/kind documentation

add manifest and artifact en doc

Fixes #

```
```

```release-note

```

```docs

```

### DIFF
--- a/cmd/ctl/artifact/artifact.go
+++ b/cmd/ctl/artifact/artifact.go
@@ -37,7 +37,7 @@ func NewCmdArtifact() *cobra.Command {
 	o := NewArtifactOptions()
 	cmd := &cobra.Command{
 		Use:   "artifact",
-		Short: "manage KubeKey artifact",
+		Short: "Manage a KubeKey offline installation package",
 	}
 
 	o.CommonOptions.AddCommonFlag(cmd)
--- a/cmd/ctl/artifact/export.go
+++ b/cmd/ctl/artifact/export.go
@@ -47,7 +47,7 @@ func NewCmdArtifactExport() *cobra.Command {
 	o := NewArtifactExportOptions()
 	cmd := &cobra.Command{
 		Use:   "export",
-		Short: "export a KubeKey artifact",
+		Short: "Export a KubeKey offline installation package",
 		Run: func(cmd *cobra.Command, args []string) {
 			util.CheckErr(o.Complete(cmd, args))
 			util.CheckErr(o.Validate(args))
--- a/cmd/ctl/create/manifest.go
+++ b/cmd/ctl/create/manifest.go
@@ -49,7 +49,7 @@ func NewCmdCreateManifest() *cobra.Command {
 	o := NewCreateManifestOptions()
 	cmd := &cobra.Command{
 		Use:   "manifest",
-		Short: "Create cluster artifact manifest file",
+		Short: "Create an offline installation package configuration file",
 		Run: func(cmd *cobra.Command, args []string) {
 			util.CheckErr(o.Complete(cmd, args))
 			util.CheckErr(o.Run())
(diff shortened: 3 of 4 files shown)


Patrick-LuoYu:
Maybe we do not need to introduce the `artifact` and `manifest` concepts, which are difficult to understand for new users. Just call them `offline installation package` and `offline installation package configuration file`.

24sama:
> Maybe we do not need to introduce the `artifact` and `manifest` concepts, which are difficult to understand for new users. Just call them `offline installation package` and `offline installation package configuration file`.

You're the definition of the `artifact` and `manifest` is correct. But the `artifact` and `manifest` is already a specific field (or a feature?) of the KubeKey, just like the `pod` of Kubernetes. So I want to explain and define these fields as much as possible and have the user accept them as if they were the `pod`.

FeynmanZhou:
> > Maybe we do not need to introduce the `artifact` and `manifest` concepts, which are difficult to understand for new users. Just call them `offline installation package` and `offline installation package configuration file`.
> 
> You're the definition of the `artifact` and `manifest` is correct. But the `artifact` and `manifest` is already a specific field (or a feature?) of the KubeKey, just like the `pod` of Kubernetes. So I want to explain and define these fields as much as possible and have the user accept them as if they were the `pod`.

IMO, `artifact` and `manifest` are very commonly used in the cloud-native area, can we add a glossary part for KubeKey to explain those definitions?

Patrick-LuoYu:
Manifest should not be used as the name of a feature. It is a general word, and any configuration file can be called a manifest. Naming a feature `manifest` is like naming the feature `configuration file`, or naming fruits of a tree `fruit`. Kubernetes names one of its resource category `pod` because it adds meaning to the ordinary word `pod`. The meaning of the word `pod` in the context of Kubernetes is different from the common meanings of `pod`. However, in the context of KubeKey, a manifest is a configuration file, the same as its original meaning, nothing else.

24sama:
> > > Maybe we do not need to introduce the `artifact` and `manifest` concepts, which are difficult to understand for new users. Just call them `offline installation package` and `offline installation package configuration file`.
> > 
> > 
> > You're the definition of the `artifact` and `manifest` is correct. But the `artifact` and `manifest` is already a specific field (or a feature?) of the KubeKey, just like the `pod` of Kubernetes. So I want to explain and define these fields as much as possible and have the user accept them as if they were the `pod`.
> 
> IMO, `artifact` and `manifest` are very commonly used in the cloud-native area, can we add a glossary part for KubeKey to explain those definitions?

Looks good to me.

Patrick-LuoYu:
To be more clear, using `manifest` as the name of a KubeKey feature is like naming a newborn kid `human`. The word `human` should not be the name of anyone. There is a great difference between general words and terms.

Patrick-LuoYu:
If you name this KubeKey feature `manifest`, how do you explain the concept `manifest`? A `manifest` is a configuration file used to create an offline installation package? This is the same as "A human is a man born in California", which is non-sense. A human is any man and woman on the Earth, not specific to man and California.

24sama:
> To be more clear, using `manifest` as the name of a KubeKey feature is like naming a newborn kid `human`. The word `human` should not be the name of anyone. There is a great difference between general words and terms.

You're right. I agree with you. So, should I remove the `artifact` and `manifest` description paragraphs? And call them offline installation package and offline installation package configuration file.

Patrick-LuoYu:
Just call them `offline installation package` and `offline installation package configuration file`. Everybody knows what they are. You don't even need to explain them.

24sama:
> Just call them `offline installation package` and `offline installation package configuration file`. Everybody knows what they are. You don't even need to explain them.

OK, I will change it later.

Patrick-LuoYu:
If you think `offline installation package` and `offline installation package configuration file` are too long, you can call them `offline package` and `offline package configuration file` with explanation that the offline package is used to install Kubernetes in an offline environment.

24sama:
@Patrick-LuoYu 
I make some updates, PTAL.

PR #1142, opened 2022-03-20
Support the use of kubeadm to manage etcd and use of existing etcd

pixiake:

### What type of PR is this?
/kind feature

At present, kubekey only supports the separate installation of etcd cluster for k8s.
It should also support the use of existing etcd and etcd managed by kubeadm.

Add some parameters to the configuration file to select etcd cluster installation mode.
Support the choice of three etcd cluster types: `kubekey`, `kubeadm` and `external`.
When the type is set to external, `endpoints`、`caFile`、`certFile` and `keyFile` should be set..
If TLS authentication is not enabled for the existing etcd, `caFile`、`certFile` and `keyFile`  need not be set.

```
kind: Cluster
metadata:
  name: sample
spec:
  ...
  etcd:
    type: external   (kubekey / kubeadm / external)   # default: kubekey
    external:
      endpoints:
      - https://192.168.6.6:2379
      caFile: /pki/etcd/ca.crt
      certFile: /pki/etcd/etcd.crt
      keyFile: /pki/etcd/etcd.key
  ...     
```


pkg/bootstrap/os/module.go
```diff
@@ -99,6 +100,7 @@ func (c *ClearOSEnvironmentModule) Init() {
 		Name:     "StopETCDService",
 		Desc:     "Stop etcd service",
 		Hosts:    c.Runtime.GetHostsByRole(common.ETCD),
```
pixiake, line 102: @24sama 
I found that if the etcd group is empty, it will be panic when initializing the task. Can we optimize it?

apis/kubekey/v1alpha2/etcd_types.go
```diff
@@ -0,0 +1,48 @@
...
+	KubeKey  = "kubekey"
+	Kubeadm  = "kubeadm"
+	External = "external"
+)
+
+type EtcdCluster struct {
+	// Type of etcd cluster, can be set to 'kubekey' 'kubeadm' 'external'
+	Type string `yaml:"type" json:"type,omitempty"`
+	// ExternalEtcd describes how to connect to an external etcd cluster when type is set to external
+	External         ExternalEtcd `yaml:"external" json:"external,omitempty"`
```
24sama, line 29: What do you think about removing the `--certificates-dir` flag? I think it's more consistent to use the `certFile` in the config to specify all cases.
BTW, may need to change the name `external`.

pkg/pipelines/add_nodes.go
```diff
@@ -53,11 +54,11 @@ func NewAddNodesPipeline(runtime *common.KubeRuntime) error {
 		&kubernetes.StatusModule{},
 		&container.InstallContainerModule{},
 		&images.PullModule{Skip: runtime.Arg.SkipPullImages},
-		&etcd.PreCheckModule{},
-		&etcd.CertsModule{},
-		&etcd.InstallETCDBinaryModule{},
-		&etcd.ConfigureModule{},
-		&etcd.BackupModule{},
+		&etcd.PreCheckModule{Skip: runtime.Cluster.Etcd.Type != kubekeyapiv1alpha2.KubeKey},
+		&etcd.CertsModule{Skip: runtime.Cluster.Etcd.Type != kubekeyapiv1alpha2.KubeKey},
```
24sama, line 58: If we add master nodes and use an `external` etcd type, this module will be skipped.

apis/kubekey/v1alpha2/cluster_types.go
```diff
@@ -239,7 +232,7 @@ func (cfg *ClusterSpec) GroupHosts() map[string][]*KubeHost {
 	if len(roleGroups[Master]) == 0 && len(roleGroups[ControlPlane]) == 0 {
 		logger.Log.Fatal(errors.New("The number of master/control-plane cannot be 0"))
 	}
-	if len(roleGroups[Etcd]) == 0 {
+	if len(roleGroups[Etcd]) == 0 && cfg.Etcd.Type == KubeKey {
 		logger.Log.Fatal(errors.New("The number of etcd cannot be 0"))
```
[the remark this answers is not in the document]
pixiake: This prompt will only be triggered when using kubekey to install etcd.

pkg/core/connector/runtime.go
```diff
@@ -130,7 +130,11 @@ func (b *BaseRuntime) SetAllHosts(hosts []Host) {
 }
 
 func (b *BaseRuntime) GetHostsByRole(role string) []Host {
-	return b.roleHosts[role]
+	if _, ok := b.roleHosts[role]; ok {
```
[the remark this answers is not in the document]
pixiake: This change did not completely solve the problem. I will continue to solve this problem when I have free time.

pkg/bootstrap/os/module.go
```diff
@@ -121,10 +123,13 @@ func (c *ClearOSEnvironmentModule) Init() {
 
 	c.Tasks = []task.Interface{
 		resetNetworkConfig,
-		stopETCD,
 		removeFiles,
 		daemonReload,
 	}
+
+	if len(c.KubeConf.Cluster.Etcd.Type) == 0 || c.KubeConf.Cluster.Etcd.Type == kubekeyv1alpha2.KubeKey {
+		c.Tasks = append([]task.Interface{stopETCD}, c.Tasks...)
```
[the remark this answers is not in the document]
pixiake: Yes, I will organize all etcd related action into one task.

apis/kubekey/v1alpha2/etcd_types.go
pixiake: When I deal with the logic here, I'm also confused about why there is this parameter `--certificates-dir`. It should have been considered at that time that there might be a need to specify the certificate generation path, but now it seems that this parameter can be removed.

This structure is defined by referring to kubeadm. 
https://pkg.go.dev/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta2#Etcd

pkg/pipelines/add_nodes.go
pixiake: OK, I got it. I will adjust it here.

pkg/core/connector/runtime.go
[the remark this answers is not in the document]
24sama: Which specific line of code triggers the `panic`, I can take a look.

pkg/bootstrap/os/module.go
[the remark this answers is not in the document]
pixiake: Done

apis/kubekey/v1alpha2/etcd_types.go
pixiake: Done

pkg/pipelines/add_nodes.go
pixiake: Done

pkg/etcd/module.go
```diff
@@ -87,16 +107,43 @@ func (c *CertsModule) Init() {
...
 		generateCerts,
 		syncCertsFile,
 		syncCertsToMaster,
 	}
 }
 
+func CertsModuleForExternal(c *CertsModule) []task.Interface {
+	fetchCerts := &task.LocalTask{
+		Name:   "FetchETCDCerts",
+		Desc:   "Fetcd etcd certs",
```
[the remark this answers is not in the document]
pixiake: Done

PR #1144, opened 2022-03-22
fix #1051: Best efforts to clear ipvs rules.

### LINKED ISSUE #1051: ipvs policies still exist after deleting cluster
### What is version of KubeKey has the issue?

latest

### What is your os environment?

CentOS 7

### KubeKey config file

```yaml
any
```

### A clear and concise description of what happend.

There are some ipvs policies created by kube-proxy during cluster creation, But these ipvs policies still exist after I deleting the whole cluster created by kubekey.

### Relevant log output

```shell
16:23:51 CST success: [node-80-68]
16:23:51 CST success: [node-80-175]
16:23:51 CST success: [node-80-152]
16:23:51 CST success: [node-80-26]
16:23:51 CST [ClearOSModule] Reset os network config
16:23:51 CST stdout: [node-80-26]
Cannot find device "nodelocaldns"
16:23:51 CST stdout: [node-80-175]
Cannot find device "nodelocaldns"
16:23:51 CST stdout: [node-80-152]
Cannot find device "nodelocaldns"
16:23:51 CST stdout: [node-80-68]
Cannot find device "nodelocaldns"
16:23:51 CST success: [node-80-26]
16:23:51 CST success: [node-80-175]
16:23:51 CST success: [node-80-152]
16:23:51 CST success: [node-80-68]
16:23:51 CST [ClearOSModule] Stop etcd service
16:24:07 CST success: [node-80-26]
16:24:07 CST success: [node-80-152]
16:24:07 CST success: [node-80-68]
16:24:07 CST [ClearOSModule] Remove cluster files
16:24:11 CST success: [node-80-26]
16:24:11 CST success: [node-80-68]
16:24:11 CST success: [node-80-175]
16:24:11 CST success: [node-80-152]
16:24:11 CST [ClearOSModule] Systemd daemon reload
16:24:12 CST success: [node-80-175]
16:24:12 CST success: [node-80-26]
16:24:12 CST success: [node-80-152]
16:24:12 CST success: [node-80-68]
16:24:12 CST [UninstallAutoRenewCertsModule] UnInstall auto renew control-plane certs
16:24:13 CST success: [node-80-26]
16:24:13 CST Pipeline[DeleteClusterPipeline] execute successful
```

### Additional information

image

tpiperatgod:
### What type of PR is this?

/kind bug

Fixes #1051

```
Since `kk` works by accessing the node via ssh and then executing a shell command, it is not suitable to use the `ipvs` library for this task.

And when using shell commands to clear the IPVS rules, it may fail because the `ipvsadm` command does not exist on the node. So, we can only use our best efforts to clear the ipvs rules.
```

None

```release-note

```

```docs

```

Signed-off-by: laminar <fangtian@kubesphere.io>

### DIFF
--- a/pkg/bootstrap/os/tasks.go
+++ b/pkg/bootstrap/os/tasks.go
@@ -18,14 +18,15 @@ package os
 
 import (
 	"fmt"
+	"path/filepath"
+	"strings"
+
 	osrelease "github.com/dominodatalab/os-release"
 	"github.com/kubesphere/kubekey/pkg/bootstrap/os/repository"
 	"github.com/kubesphere/kubekey/pkg/common"
 	"github.com/kubesphere/kubekey/pkg/core/connector"
 	"github.com/kubesphere/kubekey/pkg/utils"
 	"github.com/pkg/errors"
-	"path/filepath"
-	"strings"
 )
 
 type NodeConfigureOS struct {
@@ -163,6 +164,7 @@ var (
 		"iptables -X",
 		"iptables -F -t nat",
 		"iptables -X -t nat",
+		"ipvsadm -C",
 		"ip link del kube-ipvs0",
 		"ip link del nodelocaldns",
 	}
@@ -173,7 +175,9 @@ type ResetNetworkConfig struct {
 }
 
 func (r *ResetNetworkConfig) Execute(runtime connector.Runtime) error {
-	_, _ = runtime.GetRunner().SudoCmd(strings.Join(networkResetCmds, " && "), true)
+	for _, cmd := range networkResetCmds {
+		_, _ = runtime.GetRunner().SudoCmd(cmd, true)
+	}
 	return nil
 }
 
@@ -257,13 +261,13 @@ func (o *OnlineInstallDependencies) Execute(runtime connector.Runtime) error {
 	case "deb":
 		if _, err := runtime.GetRunner().SudoCmd(
 			"apt update;"+
-				"apt install socat conntrack ipset ebtables chrony -y;",
+				"apt install socat conntrack ipset ebtables chrony ipvsadm -y;",
 			false); err != nil {
 			return err
 		}
 	case "rpm":
 		if _, err := runtime.GetRunner().SudoCmd(
-			"yum install openssl socat conntrack ipset ebtables chrony -y",
+			"yum install openssl socat conntrack ipset ebtables chrony ipvsadm -y",
 			false); err != nil {
 			return err
 		}
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ KubeKey can install Kubernetes and KubeSphere together. The dependency that need
 | `conntrack` | Required                  | Optional but recommended  |
 | `ebtables`  | Optional but recommended  | Optional but recommended  |
 | `ipset`     | Optional but recommended  | Optional but recommended  |
+| `ipvsadm`   | Optional but recommended  | Optional but recommended  |
 
 * Networking and DNS requirements:
   * Make sure the DNS address in `/etc/resolv.conf` is available. Otherwise, it may cause some issues of DNS in cluster.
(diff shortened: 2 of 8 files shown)


tpiperatgod:
Should we use `ipvsadm -C` to clear all rules, or should we be more precise like the following script?

```shell
for address in $(ip -4 -o addr show kube-ipvs0 | awk '{print $4}' | cut -d "/" -f 1); do
  for port in 53 443 9153; do
    if [ $port -eq 53 ]; then
      ipvsadm -D -u $address:$port
    fi
    ipvsadm -D -t $address:$port
  done
done
```

pkg/bootstrap/os/tasks.go
```diff
@@ -163,6 +163,7 @@ var (
 		"iptables -X",
 		"iptables -F -t nat",
 		"iptables -X -t nat",
+		"ipvsadm -C",
```
pixiake, line 166: `ipvsadm` may not be installed on the node.

tpiperatgod: I understand, so I call it "best-effort"...😥

Or do we just need to show some tips?

pixiake: I think we can handle it this way at present. 
If we add a prompt, we need to check `ipvsadm` first on nodes.

pixiake:
> Should we use `ipvsadm -C` to clear all rules, or should we be more precise like the following script?
> 
> ```shell
> for address in $(ip -4 -o addr show kube-ipvs0 | awk '{print $4}' | cut -d "/" -f 1); do
>   for port in 53 443 9153; do
>     if [ $port -eq 53 ]; then
>       ipvsadm -D -u $address:$port
>     fi
>     ipvsadm -D -t $address:$port
>   done
> done
> ```

Although this is more precise , it should be difficult to cover every service in the cluster, right?

tpiperatgod:
> > Should we use `ipvsadm -C` to clear all rules, or should we be more precise like the following script?
> > ```shell
> > for address in $(ip -4 -o addr show kube-ipvs0 | awk '{print $4}' | cut -d "/" -f 1); do
> >   for port in 53 443 9153; do
> >     if [ $port -eq 53 ]; then
> >       ipvsadm -D -u $address:$port
> >     fi
> >     ipvsadm -D -t $address:$port
> >   done
> > done
> > ```
> 
> Although this is more precise , it should be difficult to cover every service in the cluster, right?

You're right.

pkg/bootstrap/os/tasks.go
tpiperatgod: > I think we can handle it this way at present. If we add a prompt, we need to check `ipvsadm` first on nodes.

Do you mean we don't need to handle the leftover IPVS rules? (I think that's also reasonable.)

pixiake: Just like this pr, I think we just need to add this command here.
Or we can also help users install `ipvsadm`, if they choose to use kk to install dependent software. 
https://github.com/kubesphere/kubekey/blob/37dd9a43502597fd4afd442ec5efc76d2fb05680/pkg/bootstrap/os/repository/repository_rpm.go#L89
https://github.com/kubesphere/kubekey/blob/37dd9a43502597fd4afd442ec5efc76d2fb05680/pkg/bootstrap/os/repository/repository_deb.go#L81

tpiperatgod: Follow your advice

hack/gen-repository-iso/dockerfile.centos7
```diff
@@ -9,7 +9,7 @@ RUN mkdir ${DIR} && cd ${DIR} && \
     yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo && \
     yumdownloader --resolve --destdir ./ docker-ce-${DOCKERVERSION} docker-ce-cli-${DOCKERVERSION} containerd.io \
         nss nss-sysinit nss-tools lz4 chrony bind-utils nfs-utils net-tools ipvsadm ceph-common glusterfs-fuse yum-utils \
-        curl openssl socat conntrack ebtables ipset telnet lvm2 keepalived haproxy psmisc tcpdump && \
+        curl openssl socat conntrack ebtables ipset ipvsadm telnet lvm2 keepalived haproxy psmisc tcpdump && \
```
pixiake, line 12: `ipvsadm` is already included in these three dockerfile. 😉

tpiperatgod: removed the redundant parts...

pixiake:
@all-contributors please add @tpiperatgod for code

PR #1166, opened 2022-03-31
support private registry custom domain, add to node hosts

yayuntian:
### What type of PR is this?
/kind feature

when use exists inner harbor registry custom domain, reduce manual config node hosts

add `privateRegistryIp` field in cluster registry config，when `GenerateHosts` auto add  private registry hosts

```
  registry:
    type: harbor
    plainHTTP: false
    privateRegistry: harbor.ssl.example.com
    privateRegistryIp: 192.168.41.12            # option
    auths:
      "harbor.ssl.example.com":
         username: admin
         password: Harbor12345
    namespaceOverride: ""
    registryMirrors: []
    insecureRegistries: []

```

### DIFF
--- a/apis/kubekey/v1alpha2/cluster_types.go
+++ b/apis/kubekey/v1alpha2/cluster_types.go
@@ -175,6 +175,7 @@ type RegistryConfig struct {
 	RegistryMirrors    []string             `yaml:"registryMirrors" json:"registryMirrors,omitempty"`
 	InsecureRegistries []string             `yaml:"insecureRegistries" json:"insecureRegistries,omitempty"`
 	PrivateRegistry    string               `yaml:"privateRegistry" json:"privateRegistry,omitempty"`
+	PrivateRegistryIp  string               `yaml:"privateRegistryIp" json:"privateRegistryIp,omitempty"`
 	NamespaceOverride  string               `yaml:"namespaceOverride" json:"namespaceOverride,omitempty"`
 	PlainHTTP          bool                 `yaml:"plainHTTP" json:"plainHTTP,omitempty"`
 	Auths              runtime.RawExtension `yaml:"auths" json:"auths,omitempty"`
--- a/docs/config-example.md
+++ b/docs/config-example.md
@@ -68,6 +68,7 @@ spec:
     registryMirrors: []
     insecureRegistries: []
     privateRegistry: ""
+    privateRegistryIp: ""   # if use inner private registry, auto add domain to hosts
     namespaceOverride: ""
     auths: # if docker add by `docker login`, if containerd append to `/etc/containerd/config.toml`
       "registry-1.docker.io":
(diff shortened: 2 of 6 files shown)


docs/registry.md
```diff
@@ -52,10 +52,11 @@ KubeKey now allows users to create a container image registry ([Docker registry]
        ## `harbor` can be also set for type.
        # type: "harbor"  
        privateRegistry: dockerhub.kubekey.local
+       privateRegistryIp: ""
        auths:
          "dockerhub.kubekey.local":
-           username: admin
-           password: Harbor12345
+            username: admin
```
24sama, line 58: The same as above.

docs/config-example.md
```diff
@@ -68,11 +68,12 @@ spec:
     registryMirrors: []
     insecureRegistries: []
     privateRegistry: ""
+    privateRegistryIp: ""   # if use inner private registry, auto add domain to hosts
     namespaceOverride: ""
     auths: # if docker add by `docker login`, if containerd append to `/etc/containerd/config.toml`
       "registry-1.docker.io":
-        username : "xxx"
-        password : "***"
+         username : "xxx"
```
24sama, line 75: These 2 lines may don't need to change.

24sama, line 75: These 2 lines may don't need to change.

apis/kubekey/v1alpha2/cluster_types.go
```diff
@@ -175,6 +175,7 @@ type RegistryConfig struct {
 	RegistryMirrors    []string             `yaml:"registryMirrors" json:"registryMirrors,omitempty"`
 	InsecureRegistries []string             `yaml:"insecureRegistries" json:"insecureRegistries,omitempty"`
 	PrivateRegistry    string               `yaml:"privateRegistry" json:"privateRegistry,omitempty"`
+	PrivateRegistryIp  string               `yaml:"privateRegistryIp" json:"privateRegistryIp,omitempty"`
```
24sama, line 178: PrivateRegistryIp -> PrivateRegistryIP

24sama, line 178: PrivateRegistryIp -> PrivateRegistryIP

24sama, line 178: Before,  the value of `privateRegistry` can also be set to an IP address. Would this field be ambiguous？

docs/registry.md
yayuntian: yaml unit 2 space align

docs/config-example.md
yayuntian: yaml unit 2 space align

apis/kubekey/v1alpha2/cluster_types.go
yayuntian: The `privateRegistry ` is an optional field, and the registry recommends using the domain to configure，use IP will have many permission problems

Maybe the  field  name needs to be adjusted

docs/registry.md
24sama: `runtime.RawExtension` is used here. And both of these cases can be parsed successfully.

https://github.com/kubernetes/apimachinery/blob/v0.23.5/pkg/runtime/types.go#L94

yayuntian: Just about code style, Compliant with yaml standard，will be back

PR #1168, opened 2022-04-01
support private registry custom domain, add to node hosts

yayuntian:
### What type of PR is this?
/kind feature

when use exists inner harbor registry custom domain, reduce manual config node hosts

add `privateRegistryIp` field in cluster registry config，when `GenerateHosts` auto add  private registry hosts

```
  registry:
    type: harbor
    plainHTTP: false
    privateRegistry: harbor.ssl.example.com
    privateRegistryIp: 192.168.41.12            # option
    auths:
      "harbor.ssl.example.com":
         username: admin
         password: Harbor12345
    namespaceOverride: ""
    registryMirrors: []
    insecureRegistries: []

```

### DIFF
--- a/apis/kubekey/v1alpha2/cluster_types.go
+++ b/apis/kubekey/v1alpha2/cluster_types.go
@@ -175,6 +175,7 @@ type RegistryConfig struct {
 	RegistryMirrors    []string             `yaml:"registryMirrors" json:"registryMirrors,omitempty"`
 	InsecureRegistries []string             `yaml:"insecureRegistries" json:"insecureRegistries,omitempty"`
 	PrivateRegistry    string               `yaml:"privateRegistry" json:"privateRegistry,omitempty"`
+	PrivateRegistryIp  string               `yaml:"privateRegistryIp" json:"privateRegistryIp,omitempty"`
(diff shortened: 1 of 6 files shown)


24sama:
Hi @yayuntian 
Thanks for your contribution.

What do you think of adding an additional step to do some users' customized tasks?

yayuntian:
> Hi @yayuntian Thanks for your contribution.
> 
> What do you think of adding an additional step to do some users' customized tasks?

@24sama  not agree， when the situation of privatization deployment. If it is not a general capability, it may not be merged

24sama:
> @24sama not agree， when the situation of privatization deployment. If it is not a general capability, it may not be merged


There is just an idea in my mind. Maybe we can add a kk command that can execute one single linux command or a shell script on the several nodes which is corresponding the user specified roles (e.g. etcd, master, worker or other extra customized roles).

For example:
```
kk exec "cp /root/text /home/ubuntu/text" --roles master worker
```
That will do a cp command on the master and worker nodes.

Just an idea.

yayuntian:
> 

good idea，I'll try to make it happen when I have time

PR #1169, opened 2022-04-01
Remove zombie process crontab

yuzhiquan:
### What type of PR is this?

/kind bug

Remove zombie process crontab, in some situation, this cron will kill dockershim and make pod recreated

Fixes #

```
```

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/bootstrap/os/templates/init_script.go
+++ b/pkg/bootstrap/os/templates/init_script.go
@@ -18,8 +18,9 @@ package templates
 
 import (
 	"fmt"
-	"github.com/kubesphere/kubekey/pkg/bootstrap/registry"
 	"text/template"
+
+	"github.com/kubesphere/kubekey/pkg/bootstrap/registry"
 
 	"github.com/kubesphere/kubekey/pkg/common"
 	"github.com/kubesphere/kubekey/pkg/core/connector"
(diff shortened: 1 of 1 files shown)


pixiake:
Can you describe the situation in detail? Thanks
`dockershim` was killed by this cronjob, is it because the docker service is not normal?

yuzhiquan:
> Can you describe the situation in detail? Thanks `dockershim` was killed by this cronjob, is it because the docker service is not normal?

If some pod in cluster create zombie process,and doesnot handle it,  then grep ppid will lead to the dockershim process, that's happening in our online cluster!

pixiake:
> dockershim

Can we get around it by adding filter conditions to this command? 
such as: `ps -A -ostat,ppid,cmd | grep -v *shim* | grep -e '^[Zz]' | awk '{print $2}' | xargs kill -HUP > /dev/null 2>&1"`

yuzhiquan:
> Can we get around it by adding filter conditions to this command? such as: `ps -A -ostat,ppid,cmd | grep -v *shim* | grep -e '^[Zz]' | awk '{print $2}' | xargs kill -HUP > /dev/null 2>&1"`

May be the fix can avoid this issue, but z process does not affect in many situation, so i'm not sure if flannel or dns pod make z process, for example, this cmd can handle these?

pixiake:
> > Can we get around it by adding filter conditions to this command? such as: `ps -A -ostat,ppid,cmd | grep -v *shim* | grep -e '^[Zz]' | awk '{print $2}' | xargs kill -HUP > /dev/null 2>&1"`
> 
> May be the fix can avoid this issue, but z process does not affect in many situation, so i'm not sure if flannel or dns pod make z process, for example, this cmd can handle these?

Yes, some zombie processes may be killed by mistake.

PR #1236, opened 2022-04-22
fix:  when use docker runtime , the `kk` check cgroup driver  wrong at the docker already deployed node

tanguofu:
if use docker as container  runtime,  the cgroup driver between `docker` and `kubelet` must be same.

### DIFF
--- a/pkg/kubernetes/templates/v1beta2/kubeadm_config.go
+++ b/pkg/kubernetes/templates/v1beta2/kubeadm_config.go
@@ -311,10 +311,12 @@ func GetKubeletCgroupDriver(runtime connector.Runtime, kubeConf *common.KubeConf
 	if err != nil {
 		return "", errors.Wrap(errors.WithStack(err), "Failed to get container runtime cgroup driver.")
 	}
-	if strings.Contains(checkResult, "systemd") || !strings.Contains(checkResult, "false") {
+	if strings.Contains(checkResult, "systemd") || strings.Contains(checkResult, "SystemdCgroup = true") {
 		kubeletCgroupDriver = "systemd"
-	} else {
+	} else if strings.Contains(checkResult, "cgroupfs") || strings.Contains(checkResult, "SystemdCgroup = false") {
 		kubeletCgroupDriver = "cgroupfs"
+	} else {
+		return "", errors.Errorf("Failed to get container runtime cgroup driver from %s by run %s", checkResult, cmd)
 	}
 	return kubeletCgroupDriver, nil
 }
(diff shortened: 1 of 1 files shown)


pkg/kubernetes/templates/v1beta2/kubeadm_config.go
```diff
@@ -311,10 +311,12 @@ func GetKubeletCgroupDriver(runtime connector.Runtime, kubeConf *common.KubeConf
 	if err != nil {
 		return "", errors.Wrap(errors.WithStack(err), "Failed to get container runtime cgroup driver.")
 	}
-	if strings.Contains(checkResult, "systemd") || !strings.Contains(checkResult, "false") {
+	if strings.Contains(checkResult, "systemd") || strings.Contains(checkResult, "SystemdCgroup = true") {
```
24sama, line 314: The result of this command is:
```
$ containerd config dump | grep SystemdCgroup | awk -F'= ' '{ print $2; }'
true

$ containerd config dump | grep SystemdCgroup | awk -F'= ' '{ print $2; }'
false
```
So this case `strings.Contains(checkResult, "SystemdCgroup = true")` will always be `false`.

24sama:
Hi @tanguofu 
I left some comments, PTAL.

pkg/kubernetes/templates/v1beta2/kubeadm_config.go
tanguofu: got it let me fix

24sama:
> > Hi @tanguofu Thanks for your contribution, and I left some comments, PTAL.
> 
> i remove the `awk` , because
> 
> 1. the judge can be check  explicitly such as  `SystemdCgroup = true`.
> 2. the `grep` had filter the cgroup setting.
> 
> @24sama please review , thanks.

Agree.

yjqg6666:
$  cat /etc/containerd/config.toml

```
version = 2
[plugins]
  [plugins."io.containerd.grpc.v1.cri"]
    [plugins."io.containerd.grpc.v1.cri".cni]
      bin_dir = "/usr/lib/cni"
      conf_dir = "/etc/cni/net.d"
  [plugins."io.containerd.internal.v1.opt"]
    path = "/var/lib/containerd/opt"
  [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
    [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
      SystemdCgroup = true
```
$ containerd config dump |grep -i system
`    systemd_cgroup = false`
$ containerd --version
`containerd github.com/containerd/containerd 1.4.13~ds1 1.4.13~ds1-1~deb11u4`



It may related to containerd/containerd#4900

yjqg6666:
I configured as [this comment](https://github.com/containerd/containerd/issues/4900#issuecomment-1317569507).
`$ systemctl restart containerd`
`$containerd config dump |grep -i system`
> systemd_cgroup = true
I think the check should be updated.

PR #1257, opened 2022-05-05
fix: fix docker login failure when auths username or password is empty which lead to create cluster failure

hellocn9:
fix: fix docker login failure when auths username or password is empty which lead to create cluster failure

### What type of PR is this?
/kind bug

Fix docker login failure when auths username or password is empty which lead to create cluster failure.
When auths username or password is empty, continue to do next repository's auth login. It means to ignore this repositry's auth.

Fixes #

### DIFF
--- a/pkg/container/docker.go
+++ b/pkg/container/docker.go
@@ -83,7 +83,9 @@ func (p *DockerLoginRegistry) Execute(runtime connector.Runtime) error {
 	auths := registry.DockerRegistryAuthEntries(p.KubeConf.Cluster.Registry.Auths)
 
 	for repo, entry := range auths {
-
+		if len(entry.Username) == 0 || len(entry.Password) == 0 {
+			continue
+		}
 		cmd := fmt.Sprintf("docker login --username \"%s\" --password \"%s\" %s", entry.Username, entry.Password, repo)
 		if _, err := runtime.GetRunner().SudoCmd(cmd, false); err != nil {
 			return errors.Wrapf(err, "login registry failed, cmd: %v, err:%v", cmd, err)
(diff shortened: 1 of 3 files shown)


24sama:
Hi @hellocn9 
Thanks for your contribution!

And I think you can remove this field and make `manifest-list` to use `auth.SkipTLSVerify`:
https://github.com/kubesphere/kubekey/blob/d2703d4b26e4f737b28ed03cf504fa50b613095b/apis/kubekey/v1alpha2/cluster_types.go#L180
https://github.com/kubesphere/kubekey/blob/d2703d4b26e4f737b28ed03cf504fa50b613095b/pkg/images/tasks.go#L335

pkg/container/docker.go
```diff
@@ -83,7 +83,9 @@ func (p *DockerLoginRegistry) Execute(runtime connector.Runtime) error {
 	auths := registry.DockerRegistryAuthEntries(p.KubeConf.Cluster.Registry.Auths)
 
 	for repo, entry := range auths {
-
+		if len(entry.Username) == 0 || len(entry.Password) == 0 {
```
24sama, line 86: If all elements of the array are `continue`, then the `docker login` command will not be executed and the `$HOME/.docker/config.json` file will not be generated. And the following comand `mkdir -p /.docker && cp -f $HOME/.docker/config.json /.docker/ && chmod 0644 /.docker/config.json` will report a "Not Found" error.

24sama (changes requested):

pkg/container/docker.go
hellocn9: add judgement for $HOME/.docker/config.json

24sama:
@hellocn9 
Thanks again!  BTW, please squash your commits, thank you.

PR #1307, opened 2022-06-01
Add support for missing Kubernetes versions v1.21.0 and up

### LINKED ISSUE #1306: Support Kubernetes v1.24.1 and other new versions

v2.1.1

### Describe this feature

Kubernetes recently released a new round of updates on May 24th, 2022:

- v1.21.13
- v1.22.10
- v1.23.7
- v1.24.1

Add the ability to use these versions with the --with-kubernetes flag. Currently they cannot be used as the sha256 checksum and version is not in the application.

### Additional information

_No response_

brandan-schmitz:
### What type of PR is this?

/kind feature

Adds support for all missing Kubernetes versions starting with v1.21.0 and up. 

Fixes #1306 

```
None
```

```release-note
None
```

```docs

```


24sama:
Thanks for your contribution! @brandan-schmitz 

What a great PR. This can strongly increase the number of Kubernetes versions supported by KubeKey.

LinuxSuRen (commented):
hi @brandan-schmitz , thanks for your effort. This is really great. But I'm wondering if it is possible to update the metadata automatically?

brandan-schmitz:
I would think it would be possible with some work. Are you thinking it would be built into the application and occur when a command is run or as part of a build/release process?

If part of a release process I imagine it would have a stage in a GH action where it executes a bash script or something similar that reaches out to an API or page somewhere to get a list of the kubernetes versions. From there it can automate the process of using the version numbers to find the checksum and add it into the appropriate files with sed or something similar.

If part of the application itself when a command is run, the best thing I can think of is removing the hard-coded checksum values and simply having the application fetch the .sha256 file and read the value from that based on the specified version similar to how it downloads the actual binaries.

If you have other ideas feel free to voice them as this is all I can think of on a short notice. Though this discussion could likely be moved to a new issue ticket to better track it.

24sama:
> If part of the application itself when a command is run, the best thing I can think of is removing the hard-coded checksum values and simply having the application fetch the .sha256 file and read the value from that based on the specified version similar to how it downloads the actual binaries.

Agree with that. I think this can be a feature in the kk next major or minor version. And the kk history is, users use kk to air-gapped install a Kubernetes is very complicated in the before. So, we have to hardcode the sha256 in the kk binary to reduce the files that needed to be prepared for air-gapped installation.

But kk can build an [air-gapped package](https://github.com/kubesphere/kubekey/blob/master/docs/manifest_and_artifact.md) itself and use it to create a Kubernetes cluster now. I think we can refactor this part as you said. Certainly, this needs a new issue to track.

And before the code refactoring, this PR can solve our current problem.

24sama:
@all-contributors please add @brandan-schmitz for code

Issue #997, opened 2022-01-12
Support for configuring docker's data-root


cumirror:
### Your current KubeKey version

version.BuildInfo{Version:"1.1.0", GitCommit:"70671f7", GitTreeState:"", GoVersion:"go1.14.15"}

### Describe this feature

In a production environment, user often mount additional disk for docker to use to avoid putting pressure on the system disk.
After kk installs the k8s cluster, docker uses the default /var/lib/docker as the root directory. 
Hope kk can provide configuration options to enable the data-root configuration for docker.

### Describe the solution you'd like

At present, kk can set parameters such as mirror in RegistryConfig. Maybe we can extend this structure to support data-root. When rendering the daemon.json template, set it to the docker configuration.

24sama:
Thanks for your idea. We will consider your suggestion.

PR #1375, opened 2022-07-13
support docker's data-root config

wenwutang1:
### What type of PR is this?
/kind feature

This pr is to provide  data-root config when cni is docker and delete node add timeout for drain node
feature[#997](https://github.com/kubesphere/kubekey/issues/997)

```
```

```
```
Additional documentation, usage docs, etc.:
```
```

### DIFF
--- a/pkg/kubernetes/module.go
+++ b/pkg/kubernetes/module.go
@@ -386,7 +386,7 @@ func (d *DeleteKubeNodeModule) Init() {
 		Hosts:   d.Runtime.GetHostsByRole(common.Master),
 		Prepare: new(common.OnlyFirstMaster),
 		Action:  new(DrainNode),
-		Retry:   5,
+		Retry:   2,
 	}
 
 	deleteNode := &task.RemoteTask{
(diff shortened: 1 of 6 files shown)


pkg/kubernetes/module.go
```diff
@@ -381,12 +381,13 @@ func (d *DeleteKubeNodeModule) Init() {
...
-		Prepare: new(common.OnlyFirstMaster),
-		Action:  new(DrainNode),
-		Retry:   5,
+		Name:        "DrainNode",
+		Desc:        "Node safely evict all pods",
+		Hosts:       d.Runtime.GetHostsByRole(common.Master),
+		Prepare:     new(common.OnlyFirstMaster),
+		Action:      new(DrainNode),
+		Retry:       2,
+		IgnoreError: true,
```
24sama, line 390: The functionality of this `IgnoreError` field is not yet implemented. So if there are some errors, the result will not be what you expected.

24sama:
Hi, @wenwutang1 Thanks for your contribution!
And can you leave some description about this PR? In the future, if there is anybody finds this PR, they may be confused.

pkg/kubernetes/module.go
wenwutang1: We want to ignore the error. Can we reserve it

24sama: Just do not return the error in the `Action`.
Example:
```
_, _ := runtime.GetRunner().SudoCmd("xxxx")
```

wenwutang1: ok,deleted IgnoreError

wenwutang1: ok,I deleted this field

pixiake:
@all-contributors please add @wenwutang1 for code

PR #1398, opened 2022-07-22
feat: add artifact import for phase run

### LINKED ISSUE #1389: os2022: artifact import of update_cluster phase run

latest+unreleased

### Describe this feature

For the pipeline of update_cluster,kk should have phase run support to reduce losses from duplicate implementations.

Phase run means we should divide the origin pipeline into several parts, and the second step is to import the artifact if user need. 

### Additional information

_No response_

77yu77:
### What type of PR is this?

/kind feature

This pr is to divide the artifact import from origin pipelines, which can be an independent phase cmd for phase run functions to use.

Fixes #1389

```
1. I think artifact import should independent from those pipelines which will have phase run functions, so I put it in the artifact cmds which just use a more cmd "import" to achieve it.
2. Now the "phase.go" has no more cmd, but it can make the next work easily add other phase run cmds. So I decide to add it in this pr. 

```

```release-note
It provide user to use  " artifact import -g gzip_name " to import the artifact if needed as one step of the phase run. 
```

```docs
Usage:
  kk artifact import [flags]

Flags:
      --debug              Print detailed information
  -g, --gzip string        Path to a artifact gzip
  -h, --help               help for import
```

### DIFF
--- a/cmd/ctl/alpha/artifact.go
+++ b/cmd/ctl/alpha/artifact.go
@@ -1,0 +1,74 @@
+/*
+Copyright 2022 The KubeSphere Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package alpha
+
+import (
+	"errors"
+
+	"github.com/kubesphere/kubekey/cmd/ctl/options"
+	"github.com/kubesphere/kubekey/cmd/ctl/util"
+	"github.com/kubesphere/kubekey/pkg/alpha"
+	"github.com/kubesphere/kubekey/pkg/common"
+	"github.com/spf13/cobra"
+)
+
+type ArtifactImportOptions struct {
+	CommonOptions *options.CommonOptions
+	Artifact      string
+}
+
+func NewArtifactImportOptions() *ArtifactImportOptions {
+	return &ArtifactImportOptions{
+		CommonOptions: options.NewCommonOptions(),
+	}
+}
+
+// NewCmdArtifactImport creates a new artifact import command
+func NewCmdArtifactImport() *cobra.Command {
+	o := NewArtifactImportOptions()
+	cmd := &cobra.Command{
+		Use:   "import",
+		Short: "Import a KubeKey offline installation package",
+		Run: func(cmd *cobra.Command, args []string) {
+			util.CheckErr(o.Validate(args))
+			util.CheckErr(o.Run())
+		},
+	}
+
+	o.CommonOptions.AddCommonFlag(cmd)
+	o.AddFlags(cmd)
+	return cmd
+}
+
+func (o *ArtifactImportOptions) Run() error {
+	arg := common.Argument{
+		Debug:    o.CommonOptions.Verbose,
+		Artifact: o.Artifact,
+	}
+	return alpha.ArtifactImport(arg)
+}
+
+func (o *ArtifactImportOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&o.Artifact, "artifact", "a", "", "Path to a artifact gzip")
+}
+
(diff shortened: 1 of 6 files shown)


77yu77:
The second time to run:

cmd/ctl/alpha/artifact.go
```diff
@@ -0,0 +1,64 @@
...
+func (o *ArtifactImportOptions) Run() error {
+	arg := common.Argument{
+		Debug:    o.CommonOptions.Verbose,
+		Artifact: o.Artifact,
+	}
+	return alpha.ArtifactImport(arg)
+}
+
+func (o *ArtifactImportOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&o.Artifact, "gzip", "g", "", "Path to a artifact gzip")
```
24sama, line 63: Maybe the flag `gzip` is better to be the same as before. 

https://github.com/kubesphere/kubekey/blob/97ad9f358074647a1b21d7d23403b85cf70b7949/cmd/ctl/artifact/images/push.go#L102

cmd/ctl/alpha/artifact.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/kubesphere/kubekey/pkg/common"
+	"github.com/spf13/cobra"
+)
+
+type ArtifactImportOptions struct {
+	CommonOptions *options.CommonOptions
+	Artifact      string
+}
+
+func NewUpgradePrecheckOptions() *ArtifactImportOptions {
```
24sama, line 32: The function name would be better changed to `NewArtifactImportOptions`.

cmd/ctl/alpha/artifact.go
```diff
@@ -0,0 +1,64 @@
...
+	return &ArtifactImportOptions{
+		CommonOptions: options.NewCommonOptions(),
+	}
+}
+
+// NewCmdArtifactImport creates a new artifact import command
+func NewCmdArtifactImport() *cobra.Command {
```
24sama, line 39: Is there a need to add a `Validate()` or `Complete()` method to check these `args` and `flags` provided by the user?
What happened if a user doesn't provide a path to an `artifact`?

For example:
https://github.com/kubesphere/kubekey/blob/97ad9f358074647a1b21d7d23403b85cf70b7949/cmd/ctl/artifact/images/push.go#L67
https://github.com/kubesphere/kubekey/blob/97ad9f358074647a1b21d7d23403b85cf70b7949/cmd/ctl/artifact/images/push.go#L79

24sama:
Hi @77yu77 , thanks for your contribution!
There are some comments, PTAL.

cmd/ctl/alpha/artifact.go
77yu77: OK, I thought one cmd have two "artifact" words is a little strange before. But now it is more correct than what I write.

cmd/ctl/alpha/artifact.go
77yu77: Oh, I forgot to handle the error. I will use Complete() to solve it.

cmd/ctl/alpha/artifact.go
24sama: They are two cmds.
```
kk artifact images push -a artifact.tgz
kk artifact import -a artifact.tgz
```
If there are different `flags` that point to one thing, the user maybe gets confused.

cmd/ctl/alpha/artifact.go
24sama: It happens often. 🤣

77yu77:
cmd display:
kk1
   
  
kk2

cmd/ctl/alpha/artifact.go
```diff
@@ -0,0 +1,74 @@
...
+		Artifact: o.Artifact,
+	}
+	return alpha.ArtifactImport(arg)
+}
+
+func (o *ArtifactImportOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringVarP(&o.Artifact, "artifact", "a", "", "Path to a artifact gzip")
+}
+
+func (o *ArtifactImportOptions) Complete(_ *cobra.Command, _ []string) error {
```
24sama, line 69: Usually, `Complete` is used to set the default value to the `options` and it wouldn't exit the program. The `Validate` is used to check the args if it is validated. 
IMO, if we need to return an error or exit the program, we may use the `Validate`.

24sama:
Hi @77yu77 , there is a last thing. Could you please squash your commits?

77yu77:
Ok，I have finished it.

PR #1412, opened 2022-07-29
support recent docker version 20.10.17 as an alternative.

yjqg6666:
### What type of PR is this?
/kind feature

Support recent docker version as an alternative for bugfixes.


24sama:
Hi @yjqg6666, thanks for your contributions!

PR #1418, opened 2022-08-01
feat: add upgrade binary cmd for cluster

### LINKED ISSUE #1417: Add binary download and synchronization of upgrade phase run

latest+unreleased

### Describe this feature

For the pipeline of update_cluster,kk should have phase run support to reduce losses from duplicate implementations.

Phase run means we should divide the origin pipeline into several parts, and the first step to uupgrade k8s is to download and synchronize the binary.

### Additional information

_No response_

77yu77:
### What type of PR is this?

/kind feature

This pr is to divide the binary download and synchronization from origin pipelines, which can be an independent phase cmd for phase run functions to use.

Fixes #1417

```
I add "MaxK8sVersion" in the common to make sure the desired version should be higher than the max k8s version in nodes.
```

```release-note
It provide user to use  "upgrade phase binary " to download and synchroniz binary as one step of the upgrade phase run. 

```

```docs
Usage:
  kk upgrade phase binary [flags]

Flags:
      --debug                    Print detailed information
      --download-cmd string      The user defined command to download the necessary binary files. The first param '%s' is output path, the second param '%s', is the URL (default "curl -L -o %s %s")
  -f, --filename string          Path to a configuration file
  -h, --help                     help for binary
      --ignore-err               Ignore the error message, remove the host which reported error and force to continue
      --in-cluster               Running inside the cluster
      --namespace string         KubeKey namespace to use (default "kubekey-system")
      --with-kubernetes string   Specify a supported version of kubernetes
```

### DIFF
--- a/cmd/ctl/alpha/artifact.go
+++ b/cmd/ctl/alpha/artifact.go
@@ -21,7 +21,7 @@ import (
 
 	"github.com/kubesphere/kubekey/cmd/ctl/options"
 	"github.com/kubesphere/kubekey/cmd/ctl/util"
-	"github.com/kubesphere/kubekey/pkg/alpha"
+	"github.com/kubesphere/kubekey/pkg/alpha/artifact"
 	"github.com/kubesphere/kubekey/pkg/common"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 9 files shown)


77yu77:
When I create cluster with k8s v1.22.8 and  upgrade to v1.24.1, it will fail:

image

And then upgrade to v1.23.8, it will succeed:

image

After that I upgrade to v1.23.7, it will fail:

image

And the above are in the cluster created with config file.The follow are all-in-one:
 Create cluster with k8s v1.22.10 and  upgrade to v1.23.7, it will succeed:

image

And then upgrade to v1.23.8:

image

The failed examples is similar to the above.

PR #1467, opened 2022-08-23
Support kube-vip BGP mode

zhouqiu0103:
### What type of PR is this?

/kind feature

Support kube-vip BGP mode

Fixes #

```
```

```release-note

```

```docs

```

### DIFF
--- a/pkg/loadbalancer/tasks.go
+++ b/pkg/loadbalancer/tasks.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/kubesphere/kubekey/pkg/common"
 	"github.com/kubesphere/kubekey/pkg/core/action"
@@ -157,6 +158,10 @@ type GetInterfaceName struct {
 
 func (g *GetInterfaceName) Execute(runtime connector.Runtime) error {
 	host := runtime.RemoteHost()
+	if g.KubeConf.Cluster.ControlPlaneEndpoint.KubeVip.Mode == "BGP" {
+		host.GetCache().Set("interface", "lo")
+		return nil
+	}
 	cmd := fmt.Sprintf("ip route "+
 		"| grep %s "+
 		"| sed -e \"s/^.*dev.//\" -e \"s/.proto.*//\"", host.GetAddress())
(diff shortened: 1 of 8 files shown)


pkg/loadbalancer/tasks.go
```diff
@@ -182,20 +186,50 @@ func (g *GenerateKubevipManifest) Execute(runtime connector.Runtime) error {
...
+		for _, value := range hosts {
+			address := value.GetAddress()
+			if address != host.GetAddress() {
+				if BGPPeers == "" {
+					BGPPeers = fmt.Sprintf("%s:65000::false", address)
+				} else {
+					BGPPeers = fmt.Sprintf("%s,%s:65000::false", BGPPeers, address)
+				}
+			}
+		}
```
24sama, line 201:
```go
	var BGPPeersArr []string
	for _, value := range hosts {
		address := value.GetAddress()
		if address == host.GetAddress() {
			continue
		}
		BGPPeersArr = append(BGPPeersArr, fmt.Sprintf("%s:65000::false", address))
	}
	BGPPeers := strings.Join(BGPPeersArr, ",")
```

pkg/loadbalancer/tasks.go
```diff
@@ -182,20 +186,50 @@ func (g *GenerateKubevipManifest) Execute(runtime connector.Runtime) error {
...
-			"VipInterface": interfaceName,
-			"KubeVip":      g.KubeConf.Cluster.ControlPlaneEndpoint.Address,
-			"KubevipImage": images.GetImage(runtime, g.KubeConf, "kubevip").ImageName(),
-		},
-	}
-
-	templateAction.Init(nil, nil)
-	if err := templateAction.Execute(runtime); err != nil {
-		return err
+	if g.KubeConf.Cluster.ControlPlaneEndpoint.KubeVip.Mode == "BGP" {
```
24sama, line 189: It is recommended to follow the principle of early return pattern.
```go
// BGP mode
if g.KubeConf.Cluster.ControlPlaneEndpoint.KubeVip.Mode == "BGP" {
  xxxx
  return templateAction.Execute(runtime)
}

// ARP mode
xxx
return templateAction.Execute(runtime)
```

pkg/loadbalancer/tasks.go
```diff
@@ -182,20 +186,50 @@ func (g *GenerateKubevipManifest) Execute(runtime connector.Runtime) error {
...
+				} else {
+					BGPPeers = fmt.Sprintf("%s,%s:65000::false", BGPPeers, address)
+				}
+			}
+		}
+		templateAction := action.Template{
+			Template: templates.KubevipManifest,
+			Dst:      filepath.Join(common.KubeManifestDir, templates.KubevipManifest.Name()),
+			Data: util.Data{
+				"BGPMode":      true,
```
24sama, line 206: What do you think of adding a new template variable rather than using this `BGPMode` to select different string in the golang template engine?

24sama (commented):
Here are some comments, PTAL.

PR #1501, opened 2022-09-09
fix: failed to add labels to nodes

pixiake:

### What type of PR is this?
/kind bug

Add retry for adding nodes labels.

When the `internalLoadBalancer: haproxy` is used, before the `haproxy` is started, the labeling fails because kubectl cannot connect to the kube-apiserver.

Fixes #

```
```

```release-note

```

```docs

```


tahaluindo:
ok

Issue #1486, opened 2022-09-02
使用kk离线安装k8s集群时支持指定安装ISO的依赖包


harris1024:
### Your current KubeKey version

v2.2.1

### Describe this feature

- 使用kk v2.2.1离线部署k8s集群时自定义编辑manifest.yaml、引入https://github.com/kubesphere/kubekey/releases/download/v2.2.1/centos7-rpms-amd64.iso的ISO文件后，是否有可以更透明的方式支持用户自定义ISO文件和安装离线k8s集群时自定义ISO中rpm包的清单来满足不同用户对系统初始环境的需要。

### Describe the solution you'd like

- 希望可以给config.yaml增添字段，让用户自定义rpm依赖的声明，例如
```yaml
:
 system:
    ntpServers: #  The ntp servers of chrony.
      - time1.cloud.tencent.com
      - ntp.aliyun.com
      - node1 # Set the node name in `hosts` as ntp server if no public ntp servers access.
    timezone: "Asia/Shanghai"
    rpms：
      - nfs-utils
      - yum-utils
      - bind-utils
      - glusterfs-fuse
      - lz4
      - nss
      - nss-sysinit
      - nss-tools
      - conntrack-tools
```

24sama:
Thanks for your feedback!
/kind feature-request

xiaods:
welcome pr.

arugal:
Hi @harris1024 are you working on this feature yet?

harris1024:
> Hi @harris1024 are you working on this feature yet?

OK, great. Thank you very much for your contribution. Our business needs to use this function

PR #1524, opened 2022-09-24
feat: Add rpms and debs support.

kuops:

/kind feature

Fixes # [1486](https://github.com/kubesphere/kubekey/issues/1486)

```
```

```release-note

```

```docs

```

### DIFF
--- a/pkg/bootstrap/os/tasks.go
+++ b/pkg/bootstrap/os/tasks.go
@@ -404,7 +404,14 @@ func (i *InstallPackage) Execute(runtime connector.Runtime) error {
 	}
 	r := repo.(repository.Interface)
 
-	if installErr := r.Install(runtime); installErr != nil {
+	var pkg []string
+	if _, ok := r.(*repository.Debian); ok {
+		pkg = i.KubeConf.Cluster.System.Debs
+	} else if _, ok := r.(*repository.RedhatPackageManager); ok {
+		pkg = i.KubeConf.Cluster.System.Rpms
+	}
+
+	if installErr := r.Install(runtime, pkg...); installErr != nil {
 		return errors.Wrap(errors.WithStack(installErr), "install repository package failed")
 	}
 	return nil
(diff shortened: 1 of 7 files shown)


pkg/bootstrap/os/tasks.go
```diff
@@ -398,13 +398,28 @@ type InstallPackage struct {
...
+	os := release.(*osrelease.Data).ID
+	var pkg []string
+	switch strings.ToLower(os) {
+	case "ubuntu", "debian":
+		pkg = i.KubeConf.Cluster.System.Debs
+	case "centos", "rhel":
+		pkg = i.KubeConf.Cluster.System.Rpms
+	default:
+		return fmt.Errorf("unsupported operation system %s", os)
+	}
```
24sama, line 414: I guess that would not work when the os distribution is Kylin, fedora, or others. So I suggest using assert to check the repo type, and we can use this as follow:
```go
repo, ok := host.GetCache().Get("repo")
if !ok {
	return errors.New("get repo failed by host cache")
}
r := repo.(repository.Interface)

var pkg []string
if _, ok := r.(*repository.Debian); ok {
  pkg = i.KubeConf.Cluster.System.Debs
} else if _, ok := r.(*repository.RedhatPackageManager); ok {
  pkg = i.KubeConf.Cluster.System.Rpms
}
```

24sama (commented):
Thanks for your contributions! Here are some comments, PTAL!

pkg/bootstrap/os/tasks.go
kuops: thanks for your prompt reply.

24sama:
Thanks again! And, it would be better if you can create a new PR to improve the [file](https://github.com/kubesphere/kubekey/blob/master/docs/config-example.md).

harris1024:
OK, great. Thank you very much for your contribution. Our business needs to use this function

PR #1525, opened 2022-09-26
Update Kube-ovn to 1.10.6

zhouqiu0103:
### What type of PR is this?

/kind feature
- Update Kube-ovn to 1.10.6
-  Adjusting the structure of the Kube-ovn configuration file:split a simple file `network-plugin.yaml`  into three files `kube-ovn-crd.yaml`,`ovn.yaml` and `kube-ovn.yaml`.

Fixes #

```
```

```release-note

```

```docs

```

### DIFF
--- a/pkg/plugins/network/templates/kubeovn.go
+++ b/pkg/plugins/network/templates/kubeovn.go
@@ -22,8 +22,9 @@ import (
 	"github.com/lithammer/dedent"
 )
 
-var KubeOVN = template.Must(template.New("network-plugin.yaml").Parse(
-	dedent.Dedent(`---
+var (
+	KubeOvnCrd = template.Must(template.New("kube-ovn-crd.yaml").Parse(
+		dedent.Dedent(`---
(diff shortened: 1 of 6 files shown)


pkg/plugins/network/templates/kubeovn.go
```diff
@@ -1184,7 +1185,10 @@ spec:
     kind: HtbQos
     shortNames:
       - htbqos
----
+`)))
+
+	OVN = template.Must(template.New("ovn.yaml").Parse(
+		dedent.Dedent(`---
 {{ if .DpdkMode }}
```
24sama, line 1192: I see the DpdkMode has been removed on the above, so is this piece of code still needed?

zhouqiu0103: DpdkMode was in this location before the network-plugin.yaml was split

24sama: OK, I got it.

PR #1534, opened 2022-09-29
Support for custom OpenEBS base path

arugal:
### What type of PR is this?
/kind feature
/kind api-change

Support for custom OpenEBS base path

```yaml
kind: Cluster
metadata:
  name: {{ .Options.Name }}
spec:
  storage:
    openebs:
      basePath: /var/openebs
```

Fixes #

```
```

```release-note

```

```docs

```


arugal:
CI failed, I can't see the log, need help @24sama :)

cmd/kk/pkg/config/templates/cluster.go
```diff
@@ -64,6 +64,9 @@ spec:
     ## multus support. https://github.com/k8snetworkplumbingwg/multus-cni
     multusCNI:
       enabled: false
+  storage:
+    openebs:
+      basePath: /var/openebs
```
24sama, line 69: I think we just only show this feature in the [config-example](https://github.com/kubesphere/kubekey/blob/master/docs/config-example.md) because that will be confusing to users who do not deploy the `openebs`.

cmd/kk/apis/kubekey/v1alpha2/default.go
```diff
@@ -90,6 +90,7 @@ const (
 	DefaultDNSAddress           = "114.114.114.114"
 	DefaultDpdkTunnelIface      = "br-phy"
 	DefaultCNIConfigPriority    = "01"
+	DefaultOpenEBSBasePath      = "/var/openebs"
```
24sama, line 93: Suggest `/var/openebs/local/`, because that is the same path as in the previous code.

cmd/kk/pkg/plugins/storage/modules.go
```diff
@@ -55,6 +54,7 @@ func (d *DeployLocalVolumeModule) Init() {
 			Data: util.Data{
 				"ProvisionerLocalPVImage": images.GetImage(d.Runtime, d.KubeConf, "provisioner-localpv").ImageName(),
 				"LinuxUtilsImage":         images.GetImage(d.Runtime, d.KubeConf, "linux-utils").ImageName(),
+				"BasePath":                filepath.Join(d.KubeConf.Cluster.Storage.OpenEBS.BasePath, "local"),
```
24sama, line 57: IMO, people maybe don't like the data dir is not the expected one caused we always add a subpath `local` here.

24sama (commented):
Thanks for your contributions, and sorry to reply late. Here are some comments, PTAL.

cmd/kk/pkg/plugins/storage/modules.go
arugal: Done, please review

24sama: Thanks, it would be better if you can squash your commits.

arugal: image
You can select `squash and merge` when merging,and you can also [disable](https://docs.github.com/cn/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests) other options

PR #1659, opened 2022-12-14
ipv6 support

wenwenxiong:

Fixes #

```
```

```release-note

```

```docs

```

### DIFF
--- a/cmd/kk/pkg/loadbalancer/tasks.go
+++ b/cmd/kk/pkg/loadbalancer/tasks.go
@@ -165,7 +165,8 @@ func (g *GetInterfaceName) Execute(runtime connector.Runtime) error {
 	}
 	cmd := fmt.Sprintf("ip route "+
 		"| grep ' %s ' "+
-		"| sed -e \"s/^.*dev.//\" -e \"s/.proto.*//\"", host.GetAddress())
+		"| sed -e \"s/^.*dev.//\" -e \"s/.proto.*//\""+
+		"| uniq ", host.GetAddress())
 	interfaceName, err := runtime.GetRunner().SudoCmd(cmd, false)
 	if err != nil {
 		return err


cmd/kk/pkg/plugins/network/modules.go
```diff
@@ -128,7 +129,8 @@ func deployCalico(d *DeployNetworkPluginModule) []task.Interface {
 			Template: templates.CalicoNew,
 			Dst:      filepath.Join(common.KubeConfigDir, templates.CalicoNew.Name()),
 			Data: util.Data{
-				"KubePodsCIDR":            d.KubeConf.Cluster.Network.KubePodsCIDR,
+				"KubePodsV4CIDR":            strings.Split(d.KubeConf.Cluster.Network.KubePodsCIDR, ",")[0],
+				"KubePodsV6CIDR":            strings.Split(d.KubeConf.Cluster.Network.KubePodsCIDR, ",")[1],
```
pixiake, line 133: Same as the InternalAddress, if the ipv6 address is not configured for the `KubePodsCIDR`, I think it should be reported `panic`.

cmd/kk/pkg/core/connector/host.go
```diff
@@ -65,6 +66,14 @@ func (b *BaseHost) GetInternalAddress() string {
 	return b.InternalAddress
 }
 
+func (b *BaseHost) GetInternalIPv4Address() string {
+	return strings.Split(b.InternalAddress, ",")[0]
+}
+
+func (b *BaseHost) GetInternalIPv6Address() string {
```
pixiake, line 73: If the ipv6 address is not configured for the hosts, I think it should be reported `panic`.

cmd/kk/apis/kubekey/v1alpha2/cluster_types.go
```diff
@@ -135,12 +135,15 @@ func (cfg *ClusterSpec) GenerateCertSANs() []string {
 		if host.Address != cfg.ControlPlaneEndpoint.Address {
 			extraCertSANs = append(extraCertSANs, host.Address)
 		}
-		if host.InternalAddress != host.Address && host.InternalAddress != cfg.ControlPlaneEndpoint.Address {
-			extraCertSANs = append(extraCertSANs, host.InternalAddress)
+		InternalIPv4Address := strings.Split(host.InternalAddress, ",")[0]
+		InternalIPv6Address := strings.Split(host.InternalAddress, ",")[1]
```
pixiake, line 139: If the ipv6 address is not configured for the hosts, Ithink it should be reported `panic`.

cmd/kk/pkg/core/connector/host.go
```diff
@@ -65,6 +66,14 @@ func (b *BaseHost) GetInternalAddress() string {
...
+	return strings.Split(b.InternalAddress, ",")[1]
```
pixiake, line 74: Same as the above, if the ipv6 address is not configured for the hosts, I think it should be reported `panic`.

cmd/kk/pkg/plugins/network/modules.go
pixiake, line 133: Same as the `InteranlAddress`, if the ipv6 address is not configured for the `KubePodsV6CIDR`, I think it should be reported `panic`.

pixiake:
@wenwenxiong  Thanks for this contribution !  I had added some comments，I think it may be necessary to check the relevant parameters before obtaining the ipv6 address.

wenwenxiong:
if user want to support IPV6 , the should config node ipv6 address, and write ipv6 address in config-same.yaml like i provider

docs/config-example.md
```diff
@@ -5,9 +5,9 @@ metadata:
   name: sample
 spec:
```
wenwenxiong, line 6: ipv6 address set in host internalAddresses

docs/config-example.md
```diff
@@ -80,15 +80,15 @@ spec:
       ipipMode: Always  # IPIP Mode to use for the IPv4 POOL created at start up. If set to a value other than Never, vxlanMode should be set to "Never". [Always | CrossSubnet | Never] [Default: Always]
       vxlanMode: Never  # VXLAN Mode to use for the IPv4 POOL created at start up. If set to a value other than Never, ipipMode should be set to "Never". [Always | CrossSubnet | Never] [Default: Never]
       vethMTU: 0  # The maximum transmission unit (MTU) setting determines the largest packet size that can be transmitted through your network. By default, MTU is auto-detected. [Default: 0]
-    kubePodsCIDR: 10.233.64.0/18
-    kubeServiceCIDR: 10.233.0.0/18
+    kubePodsCIDR: 10.233.64.0/18,fc00::/48
```
wenwenxiong, line 83: why user want to support ipv6, but not set ipv6 pods and service cird ? so if user relly want to support ipv6, the config must set with node ipv6 addresses and (pod,service)cidr.

PR #1666, opened 2022-12-17
fix remove node check

zhangzhiqiangcs:
fix remove node check may misstake hostnames that contains master string.
If node name is master-1, it will be considered a master node and forbidden remove the node.


pixiake:
@all-contributors please add @zhangzhiqiangcs  for code

PR #1688, opened 2023-01-10
feat(kk): allow skipping OS pre-configuration tasks

NeonSludge:
### What type of PR is this?
/kind feature

This PR adds an option to prevent KubeKey from performing basic host OS pre-configuration.
This is needed in situations when the OS has already been pre-configured by some other means (like pre-built VM images, Ansible and so on). More specifically, this option currently disables the `initOS.sh` shell script and NTP servers configuration by KubeKey. This feature is obviously meant for advanced users only and is disabled by default.

```
```

```release-note
feat(kk): allow skipping OS pre-configuration tasks
```

```docs
You can now prevent KubeKey from performing various OS pre-configuration tasks (like loading kernel modules, populating /etc/hosts, setting sysctl.conf options, configuring NTP servers, etc) via the `spec.system.skipConfigureOS` boolean parameter in the `Cluster` manifest. Note that in this case you are responsible for configuring the OS on your nodes via some other methods before using KubeKey.
```


24sama:
Thanks for your contribution! And could you please explain why you added a field in the config file rather than add a command flag?

NeonSludge:
> And could you please explain why you added a field in the config file rather than add a command flag?

Sure. My rationalization is that this feature requires the user to do quite a lot before using KubeKey (like making the LB and node addresses resolvable in some way, setting up time sync, loading modules, etc.) so it shouldn't be exposed via command flags as most users will likely never need this and will instead rely on KubeKey's built-in automation.

docs/config-example.md
```diff
@@ -42,7 +42,8 @@ spec:
     #  - name: clean tmps files
     #    bash: |
     #       rm -fr /tmp/kubekey/*
-
```
24sama, line 45: I think we can remove this line, and the format will look better this way.

24sama:
> Sure. My rationalization is that this feature requires the user to do quite a lot before using KubeKey (like making the LB and node addresses resolvable in some way, setting up time sync, loading modules, etc.) so it shouldn't be exposed via command flags as most users will likely never need this and will instead rely on KubeKey's built-in automation.

LGTM. Thanks for this contribution.

docs/config-example.md
NeonSludge: Do you mean removing the `skipConfigureOS` example line entirely or just making the comment shorter?

PR #1715, opened 2023-01-30
Feat: add cni virtual device deletion

vicoooo26:
### What type of PR is this?
/kind feature

This PR allows kk to try to clean the network config created by flannel, cilium and calico when deleting a cluster.

Currently, `kubeadm reset` won't clean the legacy network config created by CNI. If we reinstall a cluster with the previous cluster config after deleting an existing one, new CNI pods may fail to start because of CNI bridge IP address conflicts. 
[k3s upstream](https://github.com/k3s-io/k3s/blob/master/install.sh#L677) now supports flannel cleanup, we need to keep synced with upstream, and add calico and cilium clean up abilities.

```
```

None

```release-note

```

```docs

```

### DIFF
--- a/cmd/kk/pkg/bootstrap/os/tasks.go
+++ b/cmd/kk/pkg/bootstrap/os/tasks.go
@@ -172,6 +172,15 @@ var (
 		"ipvsadm -C",
 		"ip link del kube-ipvs0",
 		"ip link del nodelocaldns",
+		"ip link del cni0",
+		"ip link del flannel.1",
+		"ip link del flannel-v6.1",
+		"ip link del flannel-wg",
+		"ip link del flannel-wg-v6",
+		"ip link del cilium_host",
+		"ip link del cilium_vxlan",
+		"ip -br link show | grep 'cali[a-f0-9]*' | awk -F '@' '{print $1}' | xargs -r -t -n 1 ip link del",
+		"ip netns show 2>/dev/null | grep cni- | xargs -r -t -n 1 ip netns del",
 	}
 )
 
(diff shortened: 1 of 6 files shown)


cmd/kk/pkg/bootstrap/os/tasks.go
```diff
@@ -172,6 +172,15 @@ var (
...
 		"ip link del nodelocaldns",
+		"ip link del cni0",
+		"ip link del flannel.1",
+		"ip link del flannel-v6.1",
+		"ip link del flannel-wg",
+		"ip link del flannel-wg-v6",
+		"ip link del cilium_host",
+		"ip link del cilium_vxlan",
+		"ip -br link show | grep 'cali[a-f0-9]*' | awk -F '@' '{print $1}' | xargs -r -t -n 1 ip link del",
+		"ip netns show 2>/dev/null | grep cni- | xargs -r -t -n 1 ip netns del",
```
24sama, line 183: Can you paste more information about these 2 commands? Like some documents or screenshots.

24sama (commented):
Thanks for your contribution! And there are some comments, PTAL!

cmd/kk/pkg/bootstrap/os/tasks.go
vicoooo26: Add some docs about k8s network configuration and description for `delete-cluster` and `delete-node` command. Some explanations may not be accurate, please feel free to revise them.

24sama: Thanks!

PR #1742, opened 2023-02-20
chore: modify the codes for go-staticcheck and others

kiragoo:
### What type of PR is this?

/kind cleanup

Fixes #

```
Modify the codes for go-static-check, and remove the unused field of  `tag` about `RemoteTask` which is defined in `cmd/kk/pkg/core/task/remote_task.go`.
```

```release-note
None
```

```docs

```


cmd/kk/pkg/core/task/remote_task.go
```diff
@@ -50,7 +50,6 @@ type RemoteTask struct {
 	PipelineCache *cache.Cache
 	ModuleCache   *cache.Cache
 	Runtime       connector.Runtime
-	tag           string
```
zaunist, line 53: It looks like the `tag` has be unused here, but do we really need to remove it? I noticed that the Tag field exists in some other structures and is used.

https://github.com/kubesphere/kubekey/blob/4ee8f2ca2031433d751908b06ce6f5f1da4c39a8/cmd/kk/pkg/images/images.go#L42

The tag has been used to get image tag.

https://github.com/kubesphere/kubekey/blob/4ee8f2ca2031433d751908b06ce6f5f1da4c39a8/cmd/kk/pkg/images/images.go#L54

kiragoo: As far as I understand it, the fields defined in `RemoteTask` may be the common fields that can be used in the function of `RemoteTask` such as `Execute()` for the abstract, there are no functions defined in `remote_task.go` use the `tag` field.
The field of `Tag` which is defined in `Image`, used in functions such as `Execute()` just implements the functions of `Action-Interface`, the `Tag` in `Image` has no relationship with the `tag` in `RemoteTask`.
As a result, the `Tag` defined in `Image` just meets the requirements for implementing the `Action-Interface` to realize specific needs.

kiragoo: > It looks like the `tag` has be unused here, but do we really need to remove it? I noticed that the Tag field exists in some other structures and is used.
> 
> https://github.com/kubesphere/kubekey/blob/4ee8f2ca2031433d751908b06ce6f5f1da4c39a8/cmd/kk/pkg/images/images.go#L42
> 
> The tag has been used to get image tag.
> 
> https://github.com/kubesphere/kubekey/blob/4ee8f2ca2031433d751908b06ce6f5f1da4c39a8/cmd/kk/pkg/images/images.go#L54

As far as I understand it, the fields defined in `RemoteTask` may be the common fields that can be used in the function of `RemoteTask` such as `Execute()` for the abstract, there are no functions defined in `remote_task.go` use the `tag` field.
The field of `Tag` which is defined in `Image`, used in functions such as `Execute()` just implements the functions of `Action-Interface`, the `Tag` in `Image` has no relationship with the `tag` in `RemoteTask`.
As a result, the `Tag` defined in `Image` just meets the requirements for implementing the `Action-Interface` to realize specific needs.

zaunist (approved):

cmd/kk/pkg/core/util/file.go
```diff
@@ -237,7 +237,6 @@ func Untar(src, dst string) error {
 				return err
 			}
 
-			fmt.Println(dstPath)
```
24sama, line 240: I don't know if this line is needed to remove at present. So, I suggest keeping it.

24sama (commented):
Thanks for the contribution!

cmd/kk/pkg/core/util/file.go
kiragoo: 😀, I'll refresh the pr to keep this line.

Issue #920, opened 2021-12-23
Allinone: kubectl unable to connect to the server: x509: certificate signed by unknown authority


chaunceyjiang:
### What is version of KubeKey has the issue?

master

### What is your os environment?

deepin 15.11

### KubeKey config file

```yaml
sudo ./kk create cluster --with-kubernetes v1.21.5  --yes --container-manager docker  --with-local-storage --with-kubesphere v3.2.1
```

### A clear and concise description of what happend.

Unable to connect to the server: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "kubernetes")

![image](https://user-images.githubusercontent.com/17962021/147231699-55ce9173-cc58-4861-861d-27d44f587c5a.png)

### Relevant log output

```shell
Unable to connect to the server: x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "kubernetes")
```

SFARPak:
I had similar issue installing Air-gapped installation

https://kubesphere.io/docs/installing-on-linux/introduction/air-gapped-installation/

I temporarily solved it by using:

-subj "/CN=dockerhub.alitech.local" -addext "subjectAltName = DNS:dockerhub.alitech.local"

This is complete command:
openssl req -newkey rsa:4096 -nodes -sha256 -keyout certs/domain.key -subj "/CN=dockerhub.alitech.local" -addext "subjectAltName = DNS:dockerhub.alitech.local" -x509 -days 36500 -out certs/domain.crt

24sama:
For now, a solution:
```
cp /etc/kubernetes/admin.conf $HOME/.kube/config
```

24sama:
And this issue can be a good-first-issue.

/good-first-issue

zaunist:
/assign.  
[WeOpen Star]I would like to help

zaunist:
[WeOpen Star]I would like to help

PR #1743, opened 2023-02-20
[WIP] fix: copy kubeconfig  to .kube dir

zaunist:
### What type of PR is this?

/kind bug

Fix https://github.com/kubesphere/kubekey/issues/920. When the user using kk is not the root user, copy the admin.conf file to the current user directory.

Fixes https://github.com/kubesphere/kubekey/issues/920#

```
```

```release-note
None
```

```docs

```

### DIFF
--- a/cmd/kk/pkg/bootstrap/customscripts/module.go
+++ b/cmd/kk/pkg/bootstrap/customscripts/module.go
@@ -49,4 +49,24 @@ func (m *CustomScriptsModule) Init() {
 
 		m.Tasks = append(m.Tasks, task)
 	}
+
+	m.CopyKubeConfig()
+}
+
+func (m *CustomScriptsModule) CopyKubeConfig() {
+	taskDir := fmt.Sprintf("%s-default-script", m.Phase)
+	taskName := fmt.Sprintf("Phase:%s CopyKubeConfig", m.Phase)
+	script := kubekeyapiv1alpha2.CustomScripts{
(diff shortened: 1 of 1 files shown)


cmd/kk/pkg/bootstrap/customscripts/module.go
```diff
@@ -49,4 +49,24 @@ func (m *CustomScriptsModule) Init() {
...
+
+	m.CopyKubeConfig()
+}
+
+func (m *CustomScriptsModule) CopyKubeConfig() {
+	taskDir := fmt.Sprintf("%s-default-script", m.Phase)
+	taskName := fmt.Sprintf("Phase:%s CopyKubeConfig", m.Phase)
+	script := kubekeyapiv1alpha2.CustomScripts{
+		Name: "CopyKubeConfig",
+		Bash: "cp /etc/kubernetes/admin.conf $HOME/.kube/config",
```
xiaods, line 61: you need double check the permission and owner for admin.conf.  kubekey requirements for root user to install. if you use normal user, then the design will broken and introduce many more issues abt issues.

zaunist: All official documents use `ubuntu` users... https://github.com/kubesphere/kubekey/blob/master/docs/config-example.md

Is it ok to use `sudo cp` here? Or maybe I'll test it again after a while.

PR #1759, opened 2023-03-01
fix: docker login escape special characters

fangzhengjin:
### What type of PR is this?

/kind bug

When using Harbor's robot account, the account format is similar to `robor$xxxxx`, and the special symbol `$` needs to be escaped to log in normally.

Fixes #

```
```

```release-note

```

```docs

```


pixiake:
@fangzhengjin Thanks for this contribution !
/approve
/lgtm

PR #1470, opened 2022-08-24
feat: Add support for building AlmaLinux 9.0 repository iso

fangzhengjin:
### What this PR does / why we need it:
Add support for building AlmaLinux 9.0 repository iso
### Special notes for reviewers:
```
```

### Does this PR introduced a user-facing change?

```release-note

```

### Additional documentation, usage docs, etc.:

```docs

```

PR #1808, opened 2023-04-07
feat: Add support for building Debian 11 repository iso

yjqg6666:
### What type of PR is this?

Add one of the following kinds:
/kind feature

Support Debian Stable 11,  https://wiki.debian.org/DebianReleases.

Reference: #1470


pixiake:
@yjqg6666 Thanks for this contribution !

PR #1857, opened 2023-05-30
add clean up the VXLAN interface when delete the cluster

hellocn9:
### What type of PR is this?
/kind bug

If we create cluster with VXLAN mode before，and when we reinstall the cluster with VXLAN mode again，we will meet the situation that we can not ping other nodes's pod. Because the VXLAN does not send the package to other node. This situation will occur at some calico version. 
We clean up the VXLAN interface when delete cluster, then reinstall the cluster with VXLAN mode again. The Calico VXLAN mode network works well.
So we need clean up the VXLAN interface when we delete the cluster.


littleBlackHouse:
@hellocn9 hanks for this contributions !
/lgtm
/approve

littleBlackHouse:
@all-contributors please add @hellocn9  for code

PR #1874, opened 2023-06-13
use os or io package to replace ioutil package

hellocn9:
### What type of PR is this?

/kind cleanup

ioutil Deprecated: As of Go 1.16, the same functionality is now provided by package io or package os, and those implementations should be preferred in new code.

We already use Go 1.19, so we can change to use os or io package.

Fixes #

```
```

```release-note

```

```docs

```


littleBlackHouse:
/lgtm
/approve

PR #1877, opened 2023-06-13
stop kubelet service first when remove worker node.

hellocn9:
### What type of PR is this?

/kind bug

When we remove node from cluster, we will execute "ResetNetworkConfig" and “RemoveFiles” task. At task of "RemoveFiles", it will delete files including "/etc/systemd/system/kubelet.service" and "/etc/systemd/system/kubelet.service.d". But there is no operation for stop kubelet service. So kubelet service will get some errors when execute "DaemonReload"  task last.

So, we need to stop kubelet service first, then we can execute "RemoveFiles" task.

Fixes #

```
```

```release-note

```

```docs

```


littleBlackHouse:
/lgtm
/approve

PR #1882, opened 2023-06-21
chore: remove refs to deprecated io/ioutil

testwill:
### What type of PR is this?
/kind cleanup

"io/ioutil" has been deprecated since Go 1.19: As of Go 1.16

Fixes #

```
```

```release-note
None
```

```docs

```

### DIFF
--- a/cmd/kk/pkg/registry/docker_registry_config.go
+++ b/cmd/kk/pkg/registry/docker_registry_config.go
@@ -18,7 +18,7 @@ package registry
 
 import (
 	"encoding/json"
-	"io/ioutil"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


cmd/kk/pkg/registry/docker_registry_config.go
littleBlackHouse, line 1: Thanks for this contributions !
Would it be better to modify the `os.FileInfo` parameter to `os.DirEntry` in line 125 of the `hasFile` function? This way, it would eliminate an extra loop.

testwill: Hello, hasFile needs to be modified.

littleBlackHouse: it's okay. just modified it.
The function hasFile is only used within the LookupCertsFile function, specifically to check if a certain file exists within the directory being read at line 84 fs, err := ioutil.ReadDir(absPath).

littleBlackHouse:
@all-contributors please add @testwill  for code,doc

Issue #1885, opened 2023-06-21
Change imagePullPolicy from Always to IfNotPresent for haproxyManifest


kellanfan:
### What is version of KubeKey has the issue?

3.3.1

### What is your os environment?

ubuntu 20.04

### A clear and concise description of what happend.

In ha-mode, when the environment cannot connect to the Internet, after restarting the node, the local haproxy pod will fail to start normally because the ImagePullPloicy is Always, and the node status is nodeready.
So need to change imagePullPolicy from Always to IfNotPresent in file cmd/kk/pkg/loadbalancer/templates/haproxyManifest.go

PR #1896, opened 2023-06-29
fix: Modify the imagePullPolicy to make it more practical for offline…

littleBlackHouse:
… scenarios.

### What type of PR is this?

/kind bug

In offline scenarios, it is not possible to pull images.We should modify the imagePullPolicy to "IfNotPresent" to accommodate both offline and online scenarios.

https://github.com/kubesphere/kubekey/issues/1885

there is default policy
https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy

### DIFF
--- a/cmd/kk/pkg/loadbalancer/templates/haproxyManifest.go
+++ b/cmd/kk/pkg/loadbalancer/templates/haproxyManifest.go
@@ -42,7 +42,7 @@ spec:
   containers:
   - name: haproxy
     image: {{ .HaproxyImage }}
-    imagePullPolicy: Always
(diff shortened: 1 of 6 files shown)


LinuxSuRen (commented):
I prefer to provide a config item for `imagePullPolicy`. So the users could make their own decision.

littleBlackHouse:
> I prefer to provide a config item for `imagePullPolicy`. So the users could make their own decision.

Yes your right!
This PR is made to fix this issue by uniformly setting the `imagePullPolicy`  to `IfNotPresent`. The next step will be to make it a configurable option in the cluster's configuration file

PR #1901, opened 2023-07-05
feat: add an initial stage at the very start cluster creating phase

LinuxSuRen:
### What type of PR is this?

/kind feature

In many cases, users need to install some packages before running `kk`. For instance: `chrony`, or `conntrack`.

Fixes #

```
None
```

```release-note
Add an initial step during the cluster creating phase
```

```docs
See also the example file
```

### DIFF
--- a/docs/config-example.md
+++ b/docs/config-example.md
@@ -42,7 +42,16 @@ spec:
     # Specify additional packages to be installed. The ISO file which is contained in the artifact is required.
     debs: 
       - nfs-common
-    #preInstall:  # Specify custom init shell scripts for each nodes, and execute according to the list order.
+    # initials:
+    #   - name: Install hd tool
+    #     bash: curl https://gitee.com/linuxsuren/tools/raw/master/install-zh.sh|bash
+    #   - name: Install packages
+    #     bash: |
+    #       hd setup --proxy ghproxy.com --provider gitee
+    #       hd i chrony
+    #       hd i conntrack
+    #       hd i socat
+    #preInstall:  # Specify custom init shell scripts for each nodes, and execute according to the list order before install container runtime.
     #  - name: format and mount disk  
     #    bash: /bin/bash -x setup-disk.sh
     #    materials: # scripts can has some dependency materials. those will copy to the node        
--- a/docs/config-example.md
+++ b/docs/config-example.md
@@ -42,22 +42,13 @@ spec:
     # Specify additional packages to be installed. The ISO file which is contained in the artifact is required.
     debs: 
       - nfs-common
-    # initials:
-    #   - name: Install hd tool
-    #     bash: curl https://gitee.com/linuxsuren/tools/raw/master/install-zh.sh|bash
-    #   - name: Install packages
-    #     bash: |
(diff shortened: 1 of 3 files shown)


LinuxSuRen:
See also the test output screenshots:

image

and the example config https://github.com/kubesphere/kubekey/pull/1901/files#diff-39a68052dffc6c617e1b3b29d9c91a3cc7e6239ce5ea90a812ca059bc7323776L45

LinuxSuRen:
Please [download it](https://github.com/LinuxSuRen/kubekey/releases/tag/v1.0.17) in case anyone wants to have a try.

docs/config-example.md
```diff
@@ -42,7 +42,16 @@ spec:
     # Specify additional packages to be installed. The ISO file which is contained in the artifact is required.
     debs: 
       - nfs-common
-    #preInstall:  # Specify custom init shell scripts for each nodes, and execute according to the list order.
+    # initials:
```
pixiake, line 45: I think `preInstall` may be more intuitive than `initials`, kk also supports configuring `postInstall`, and some users are already used `preInstall` in their configuration files.

LinuxSuRen: Yes, `preInstall` looks like more reasonable. But it runs after the checking. For instance: `socat` is required, but I cannot install `socat` during `preInstall`. Because it will fail before start `preInstall`.

LinuxSuRen: Any  suggestions?

pixiake: kk has a command `kk init os -f config.yaml` to install dependency packages， such as: `socat`、`conntrack` .

The original idea of `preInstall` is to support users to do some custom operations before installing kubernetes cluster. such as partitioning disks on nodes, mounting disk to the data directory of docker, and so on.

So I think we can use `kk init os -f config.yaml` to install the necessary dependency packages, and other additional operations can be configured through `preInstall`.

pixiake: If it is an offline environment, kk also supports specifying `--with-packages` to install operating system dependencies using `rpms` or `debs` in offline packages.

https://github.com/kubesphere/kubekey/blob/master/docs/manifest_and_artifact.md
https://github.com/kubesphere/kubekey/blob/master/docs/manifest-example.md

LinuxSuRen: I still believe the `preInstall` is not a real `pre` action. It's too late. An easy way to fix that is moving `preInstall` to the first (or second) stage. I just want to do something before any verifies.

pixiake: I got it. I think it‘s ok to advance the execution order of `preInstall`.

So, I think we should directly advance the execution of `preInstall`, and don't need to add this parameter `Initials`, about you?

LinuxSuRen: I'm going to do it if you agree with the solution to move `preInstall` after `GreetingsModule`.

pixiake: Yes, I think it is reasonable to do so.  Thanks !

LinuxSuRen: Done!

pixiake:
@LinuxSuRen Thanks for this contribution !
/lgtm
/approve

PR #1907, opened 2023-07-07
fix: imagefullName support more than two slash-separated

wongearl:
### What type of PR is this?
/kind bug

for imagefullName support more than two slash-separated, like 10.121.218.184:30002/cache/goharbor/notary-server-photon:v2.8.2, While the [OCI Distribution Specification](https://github.com/opencontainers/distribution-spec) supports more than two slash-separated components

Fixes #

```
```

```release-note
  None
```

```docs

```

### DIFF
--- a/cmd/kk/pkg/images/utils.go
+++ b/cmd/kk/pkg/images/utils.go
@@ -173,11 +173,18 @@ func NewManifestSpec(image string, entries []manifesttypes.ManifestEntry) manife
 
 func validateImageName(imageFullName string) error {
 	image := strings.Split(imageFullName, "/")
-	if len(image) != 3 {
-		return errors.Errorf("image %s is invalid, only the format \"registry/namespace/name:tag\" is supported", imageFullName)
+	if len(image) < 3 {
+		return errors.Errorf("image %s is invalid, image PATH need contain at least two slash-separated", imageFullName)
 	}
-	if len(strings.Split(image[2], ":")) != 2 {
-		return errors.Errorf("image %s is invalid, only the format \"registry/namespace/name:tag\" is supported", imageFullName)
+	if len(strings.Split(image[len(image)-1], ":")) != 2 {
+		return errors.Errorf(`image %s is invalid, image PATH need contain ":"`, imageFullName)
 	}
 	return nil
+}
+
+func suffixImageName(imageFullName []string) string {
+	if len(imageFullName) >= 2 {
+		return strings.Join(imageFullName, "/")
+	}
+	return imageFullName[0]
 }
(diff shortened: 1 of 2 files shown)


cmd/kk/pkg/images/utils.go
```diff
@@ -173,11 +173,18 @@ func NewManifestSpec(image string, entries []manifesttypes.ManifestEntry) manife
 
 func validateImageName(imageFullName string) error {
 	image := strings.Split(imageFullName, "/")
-	if len(image) != 3 {
-		return errors.Errorf("image %s is invalid, only the format \"registry/namespace/name:tag\" is supported", imageFullName)
+	if len(image) >= 3 {
```
LinuxSuRen, line 176: Bigger than 3 or less than 3?

cmd/kk/pkg/images/utils.go
```diff
@@ -173,11 +173,18 @@ func NewManifestSpec(image string, entries []manifesttypes.ManifestEntry) manife
...
+		return errors.Errorf("image %s is invalid,image PATH need contain  at least two slash-separated", imageFullName)
```
LinuxSuRen, line 177: Extra whitespace between `contain  at` is not necessary. A whitespace after comma is necessary. `invalid,image`

LinuxSuRen (changes requested):

cmd/kk/pkg/images/utils.go
wongearl: only bigger than 3 is legal

cmd/kk/pkg/images/utils.go
wongearl: ok

cmd/kk/pkg/images/utils.go
wongearl: i get it

cmd/kk/pkg/images/utils.go
```diff
@@ -173,11 +173,18 @@ func NewManifestSpec(image string, entries []manifesttypes.ManifestEntry) manife
...
 	}
-	if len(strings.Split(image[2], ":")) != 2 {
-		return errors.Errorf("image %s is invalid, only the format \"registry/namespace/name:tag\" is supported", imageFullName)
+	if len(strings.Split(image[len(image)-1], ":")) != 2 {
+		return errors.Errorf(`image %s is invalid, image PATH need contain ":"`, imageFullName)
 	}
 	return nil
 }
+
+func suffixImageName(imageFullName []string) string {
```
LinuxSuRen, line 185: Good to have an unit test here.

PR #1951, opened 2023-08-14
When joining a node, keep HAProxy consistent, as in the case of adding a node.

gwgrisk:
This commit is aimed at ensuring consistency between the effects of joining a node and adding a node, by enabling haproxy in both cases. Prior to this modification, haproxy was not started when joining a node, leading to functional abnormalities.


gwgrisk:
/assign @pixiake

review，please

gwgrisk:
invailid when join，

PR #1992, opened 2023-09-20
 support harbor docker-compose ha design and come true

wenwenxiong:
### What type of PR is this?

/kind feature
support harbor docker-compose ha design and come true
multi node harbor use the same certs ,the certs contain each registry role node name and IP for cert common name.
each habror can config createReplication use script
use keepalived config vip for one of registry node fail, the other registry one can provider harbor service with vip assign to this registry node.

Fixes #

```
```

```release-note

```

```docs

```


pixiake:
@wenwenxiong That's Greate ! Thanks for this contribution !

/lgtm
/approve

PR #2013, opened 2023-10-11
make docker and containerd no parent-child process relationship

wenwenxiong:
### What type of PR is this?

when use kubekey docker and multus calico meet many issue with runtime ,  i meet docker runtime  stuck and restart it in stuck too.
when i use docker-ce deb package deploy, i found it use systemd manager docker and containerd  service like this pr to do .
 i change docker runtime in that way fix the docker stuck issue.

Fixes #

```
```

```release-note

```

```docs

```


pixiake:
@wenwenxiong Thanks for this contribution!
/lgtm
/approve

PR #2038, opened 2023-11-06
fix: backup script doesn't work when set etcd dataDir to another dir except '/var/lib/etcd'

liuxu623:
### What type of PR is this?
/kind bug

backup script doesn't work when set etcd dataDir to another dir except '/var/lib/etcd'


pixiake:
@liuxu623 Thanks for this contribution !
/lgtm
/approve

pixiake:
@all-contributors please add @liuxu623  for code

PR #2041, opened 2023-11-09
fix: redundant interface for kubevip when add k8s nodes

qyz87:
get redundant interface for kubevip when add k8s nodes

### What type of PR is this?
kind bug

first deploy k8s which has only one nodes, then use kk add 2 nodes to k8s cluster, and use kube-vip for loadbalancer.  now  the firs node vip_interface contains two calico interfaces:

```
    - name: vip_interface
      value: calic3e37beb20a scope link src 192.168.122.101
cali97a236bb38c scope link src 192.168.122.101
eno1

```

   
Fixes #

```
```

```release-note

```

```docs

```


pixiake:
@qyz87 Thanks for this contribution !
/lgtm
/approve

pixiake:
@all-contributors please add @qyz87  for code

PR #2142, opened 2024-02-23
add ipv6 support in k8s and calico  config,it compatible only ipv4 provide

wenwenxiong:
### What type of PR is this?
/kind feature

add ipv6 support in k8s and calico  config,it compatible only ipv4 provide
like doc/config-example.md
just change host.internalAddress, kubePodsCIDR, kubeServiceCIDR with ipv6 address and cidr, you can deploy k8s with calico support ipv4/ipv6 cluster
and it compatible only ipv4 provide

Fixes #

```
```

```release-note

```

```docs

```


pixiake:
@wenwenxiong Thanks for this contributions !
/lgtm
/approve

PR #2194, opened 2024-04-03
Fix the issue that the cri-dockerd.sock file is not automatically deleted when stopping the cri-docker service.

hellocn9:
Fix the issue that the `/var/run/cri-dockerd.sock` file is not automatically deleted when stopping the cri-docker service, and when installing the cluster again, kk detects the existence of the `/var/run/cri-dockerd.sock` file and no longer installs the cri-docker service, which in turn leads to the failure of the cluster installation.

### What type of PR is this?
/kind bug

When stopping the cri-docker service, the `/var/run/cri-dockerd.sock` file is not automatically deleted.
When we install the cluster again, kk detects the existence of the `/var/run/cri-dockerd.sock` file and no longer installs the cri-docker service, which in turn leads to the failure of the cluster installation.

Fixes #

```
```

```release-note

```

```docs

```


pixiake:
/lgtm
/approve

PR #2205, opened 2024-04-11
Add Flag --skip-remove-artifact to remain kubekey/artifact dir after exporting an artifact

hellocn9:
Add Flag --skip-remove-artifact to remain kubekey/artifact dir after exporting an artifact

### What type of PR is this?
/kind feature

Some scenarios may require recreating artifact, such as adding a new image, which would need to re-download all files due to the deletion of the artifact directory. So, add an option to keep the artifact directory to reuse some of the files.

Fixes #

```
```

```release-note

```

```docs

```


pixiake:
/lgtm
/approve

PR #2206, opened 2024-04-12
remove the IPV6 domain mapping from the /etc/hosts file when the host does not have an IPV6 address.

hellocn9:
remove the IPV6 domain mapping from the `/etc/hosts` file when the host does not have an IPV6 address.

### What type of PR is this?
/kind cleanup

When the host does not have IPV6 address, ConfigureOSModule will also write IPV6 domain mapping. It will cause /etc/hosts like following
image
It should not be IPV6 domain mapping when the host does not have IPV6 address in `/etc/hosts`.

Fixes #

```
```

```release-note

```

```docs

```


pixiake:
/lgtm
/approve

PR #2257, opened 2024-05-26
feat: set CustomScriptTask workdir to remoteTaskHome

will4j:
### What type of PR is this?

/kind bug

when using custom scripts with materials, bash command' workdir is not at remoteTaskHome which saves the material files, so the custom scripts execute failed as it can not find the material files, change workdir to remoteTaskHome before bash command run makes it much better.

like this:
```yaml
spec:
  system:
    preInstall:
      - name: install nfs server
        role: nfs-server
        bash: |
          # workdir at remoteTaskHome
          /bin/bash -x install_nfs_server.sh
        materials:
          - ./scripts/install_nfs_server.sh
```

Fixes #

```
```

```release-note
None
```

```docs

```


pixiake:
@will4j Thanks for this contributions !
/lgtm
/approve

PR #2259, opened 2024-05-27
add check support k8s version

bzd111:
Use the following command to create a cluster: `kk create cluster -f cluster.yaml`.

If a Kubernetes version specified in cluster.yaml is not found in versions/components.json, the installation process should terminate. Instead of exiting after a failed SHA256 comparison following the download.

### What type of PR is this?
/kind feature

Optionally add one or more of the following kinds if applicable:
/kind flake

Check the supported k8s versions and terminate early in the precheck.

```
```

```release-note
no
```

### DIFF
--- a/cmd/kk/pkg/bootstrap/confirm/tasks.go
+++ b/cmd/kk/pkg/bootstrap/confirm/tasks.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"os"
 	"regexp"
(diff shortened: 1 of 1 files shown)


bzd111:
@pixiake 
Apologies, I overlooked that the Go version is 1.19, which does not support the use of built-in slices available in 1.21. 
Please review the commit.

PR #2406, opened 2024-09-14
fix: ssh formatting exception bug when executing commands

dbbDylan:
### What type of PR is this?

/kind bug

- String formatting directly by `fmt.Sprintf` may cause unexpected error when executing a single command as a result of multiple double-quote transpositions. The more safe way is to use `exec.Command` function. 
- Fix from `file://` to `file:///` in `init_repository.yaml`

For more information on release notes see: https://github.com/kubernetes/community/blob/master/contributors/guide/release-notes.md
-->
```release-note
- fix: ssh formatting exception bug when executing commands
- fix: from `file://` to `file:///` in `init_repository.yaml`
```


dbbDylan:
This way can correctly execute multi-line commands, but use one `nolint:gosec` annotation to avoid lint error.

PR #2410, opened 2024-09-18
feat: update kkcore CRDs

dbbDylan:
### What type of PR is this?

/kind feature
/kind design

Update kkcore CRDs for CAPKK

Fixes #

```release-note
- update: add `PipelineTemplate` CRD for CAPKK, used for create `Pipeline`.
- update: add `.Status.HostMachineMapping` for `Inventory`, used for update status between `KKMachine` and `Hosts`.
```


pkg/apis/core/v1/inventory_types.go
```diff
@@ -66,6 +80,33 @@ type InventoryList struct {
...
+					unavailableHosts[hostName] = struct{}{}
+					validHosts = append(validHosts, hostName)
+				}
+			}
+		}
+		hosts = append(hosts, validHosts...)
+	}
+
+	return hosts
+}
```
redscholar, line 108: move in variable package

pkg/apis/core/v1/pipelinetemplate_types.go
redscholar, line 1: should not set as a crd.

pkg/apis/core/v1/inventory_types.go
dbbDylan: This logic is unique for `inventory` CRD, so I don't need to move it.

By the way, `variable` is for `playbook`, which will store variables when the playbook being executed. But `GetHostsFromGroup` used for init a valid `inventory`.

pkg/apis/core/v1/pipelinetemplate_types.go
dbbDylan: Has been justified in PR[#2425 ](https://github.com/kubesphere/kubekey/pull/2425)

PR #2534, opened 2025-04-16
fix: Fixed kubelet's cgroup configuration always using the default systemd

YunShiHang:
### What type of PR is this?

/kind bug

Solves the problem that kubelet's cgroup configuration always uses the default systemd

Fixes #

```
```

```release-note
Affects the node kubelet config when creating a cluster or adding a new node to an existing cluster
```

```docs

```


YunShiHang:
@redscholar Please help me review the changes, thank you

PR #2752, opened 2025-09-08
bugfix : fix issue 6796

zuoxuesong-worker:
bugfix : fix bug 6796

### What type of PR is this?
/kind bug

change listip api , add return args: added
it means checked ip already added or not

Fixes #
https://github.com/kubesphere/project/issues/6796

```
None
```

```release-note
fix issue 6796
```

```docs

```

### DIFF
--- a/pkg/web/handler/resources.go
+++ b/pkg/web/handler/resources.go
@@ -190,6 +190,30 @@ func (h ResourceHandler) ListIP(request *restful.Request, response *restful.Resp
 	cidr := request.QueryParameter("cidr")
 	sshPort := query.DefaultString(request.QueryParameter("sshPort"), "22")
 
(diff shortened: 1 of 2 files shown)


pkg/web/handler/resources.go
```diff
@@ -190,6 +190,29 @@ func (h ResourceHandler) ListIP(request *restful.Request, response *restful.Resp
 	cidr := request.QueryParameter("cidr")
 	sshPort := query.DefaultString(request.QueryParameter("sshPort"), "22")
 
+	var existsInventoryList kkcorev1.InventoryList
+	err := h.client.List(request.Request.Context(), &existsInventoryList)
```
redscholar, line 194: error should not be ignored

PR #2768, opened 2025-09-18
feat : change iso installer

zuoxuesong-worker:
feat : change iso installer

### What type of PR is this?
/kind feature

change iso installer

Fixes #

```
None
```

```release-note
change iso installer
```

```docs

```


pkg/connector/helper.go
redscholar, line 1: maybe you are looking for:
https://github.com/kubesphere/kubekey/blob/f3c66de4dd5d19f2d31f03ce3930c1a0b50f7174/cmd/kk/app/options/option.go#L307-L325

PR #2816, opened 2025-10-20
Update README.md

gshk99:
A small fix in README.md - the "--ui-path" key was specified using a single hyphen.

### What type of PR is this?

A small fix in README.md 

In the README.md file, the "--ui-path" key was specified using a single hyphen.


gshk99:
will create new pull request due to changed code base

PR #2817, opened 2025-10-20
A small fix in README files - the "--ui-path" key was specified using a single hyphen.

gshk99:
### What type of PR is this?

A small fix in README files

In the README.md and README_zh-CN.md files, the "--ui-path" key was specified using a single hyphen.


gshk99:
Wrong pull request

PR #2872, opened 2025-11-26
feat: support pre_install & post_install scripts by global config

will4j:
### What type of PR is this?

/kind feature

support pre install & post install scripts for specific group, using scripts file in `scripts_dir`。
```yaml
apiVersion: kubekey.kubesphere.io/v1
kind: Config
spec:
  ...
  pre_install:
    - group: my_group_1
      script: my_scripts_1.sh
  post_install:
    - group: my_group_2
      script: my_scripts_2.sh
  ...
```
config above will execute `{{ .scripts_dir }}/my_scripts_1.sh` as pre install scripts in all hosts of group `my_group_1` and `{{ .scripts_dir }}/my_scripts_2.sh` as post install scripts in all hosts of group `my_group_2`.

to support pre_install list by `loop`，move `when` check of `Task` to loop inner, as it is in ansible ***When combining conditionals with a loop, the when: statement is processed separately for each item***. this support use `item` of loop in `when` expression, like:
```yaml
when: "{{ getStringSlice .groups .item.group | default list | has .inventory_hostname }}"
```

`get` func in sprig use `map[string]interface {}` as param, well variable `groups` is type of `map[string][]string`, template `{{ get .groups .item.group }}` failed with error msg：`expected map[string]interface {}; got map[string][]string`, that's why a `getStringSlice` func is added to support get group in templates.

Fixes #

```
```

```release-note
Support pre_install & post_install variable to config scripts of specific group
```

```docs

```

### DIFF
--- a/pkg/converter/tmpl/template_test.go
+++ b/pkg/converter/tmpl/template_test.go
@@ -257,7 +257,7 @@ func TestParseValue(t *testing.T) {
 			input: "{{ getStringSlice .foo \"foo\" }}",
 			variable: map[string]any{
 				"foo": map[string][]string{
-					"foo": []string{"bar1", "bar2"},
+					"foo": {"bar1", "bar2"},
 				},
 			},
 			excepted: []byte("[bar1 bar2]"),


pkg/converter/tmpl/template_test.go
```diff
@@ -252,6 +252,16 @@ func TestParseValue(t *testing.T) {
 			},
 			excepted: []byte("bar"),
 		},
+		{
+			name:  "get string slice from dictionary",
+			input: "{{ getStringSlice .foo \"foo\" }}",
+			variable: map[string]any{
+				"foo": map[string][]string{
+					"foo": []string{"bar1", "bar2"},
```
redscholar: To fix the gofmt lint, it should be: `"foo": {"bar1", "bar2"},`

redscholar:
@will4j Thanks for the PR. You still have an unresolved code-view comment from gemini-code-assist and a CI lint issue. Once those are addressed, we can merge it.

PR #2877, opened 2025-12-01
fix: some failed version check when create cluster

will4j:
### What type of PR is this?

/kind bug

Fix some failed version check which will cause unnecessary retries when run create cluster playbook.
1. runc version check
  `.runc_install_version.stdout` has string like `runc version 1.1.12\r\r\ncommit: v1.1.12-0-g51d5e946\r\r\nspec: 1.0.2-dev\r\r\ngo: go1.20.13\r\r\nlibseccomp: 2.5.4`,  `contains "1.1.12\n"` will not work, use `regexMatch` instead.
2. kubelet version check
  `--short` flag of kubectl was removed from kubernetes 1.28, see [this k8s pr]( https://github.com/kubernetes/kubernetes/pull/116720). command should diff by kube version to avoid **unknown flag** error. in addition, **Server Version** is not in stdout but other msg in stderr before a node joined to cluster, use **Client Version** is better.
3. helm version check
  command `helm version --template "{{ .Version }}"` will first be parsed as template by kubekey, then the result as real command executed at remote hosts, which will be `helm version --template "<no value>"` in this case, so a translate of special character `{{` and `}}` is needed.

Fixes #

```
```

```release-note
none
```

```docs

```

### DIFF
--- a/builtin/capkk/roles/install/kubernetes/tasks/main.yaml
+++ b/builtin/capkk/roles/install/kubernetes/tasks/main.yaml
@@ -1,7 +1,7 @@
 ---
 - name: Check if helm is installed
   ignore_errors: true
-  command: helm version --template "{{ .Version }}"
+  command: helm version --template "{{ "{{" }} .Version {{ "}}" }}"
   register: helm_install_version
 - name: Install helm
   when: or (.helm_install_version.error | empty | not) (.helm_install_version.stdout | ne .helm_version)


builtin/core/roles/kubernetes/pre-kubernetes/tasks/install_binaries.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 - name: Binary | Verify if Helm is already installed
   ignore_errors: true
-  command: helm version --template "{{ .Version }}"
```
redscholar, line 4: You can use a better option:
```yaml
command: |
  helm version --template "{{ .Version }}"
```

redscholar: Running `kk create cluster -v 5` will show the actual commands being executed.

redscholar: I got it

PR #2880, opened 2025-12-02
fix: failed to create local dir x permission denied

will4j:
### What type of PR is this?

/kind bug

When run command `kk create cluster` in a k8s node, copy binary to remote hosts will execute by `localConnector`, many error like *failed to copy absolute file: failed to create local dir "/tmp/etc/kubernetes/pki/etcd/ca.crt": mkdir /tmp/etc/kubernetes/pki: permission denied* occurs.

image

image

the reason is directory created by `localConnector` with src file mode, which may cause no permission to write or execute to the directory, and should be fixed by give directory with specific mode `os.ModePerm`.

Fixes #

```
```

```release-note
fix: failed to create local dir x permission denied 
```

```docs

```

### DIFF
--- a/pkg/const/common.go
+++ b/pkg/const/common.go
@@ -127,5 +127,6 @@ const (
 )
 
 const (
-	PermDirPublic = os.FileMode(0755)
+	// PermDirPublic means public permission of directory, something like os.ModePerm
+	PermDirPublic os.FileMode = 0755
 )
--- a/pkg/modules/copy.go
+++ b/pkg/modules/copy.go
@@ -217,8 +217,10 @@ func (ca copyArgs) handleRelativePath(ctx context.Context, options ExecOptions, 
 }
 
 // copyAbsoluteDir copies all files from an absolute directory to the remote host.
-func (ca copyArgs) copyAbsoluteDir(ctx context.Context, conn connector.Connector) error {
-	defer ca.ensureDestDirMode(ctx, conn)
+func (ca copyArgs) copyAbsoluteDir(ctx context.Context, conn connector.Connector) (resRrr error) {
+	defer func() {
+		resRrr = ca.ensureDestDirMode(ctx, conn)
+	}()
 	return filepath.WalkDir(ca.src, func(path string, d fs.DirEntry, err error) error {
 		// Only copy files, skip directories
 		if d.IsDir() {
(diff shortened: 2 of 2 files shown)


redscholar:
This approach isn’t ideal because the mode parameter for copy or template no longer takes effect.like
```yaml
copy:
  src: xxx
  dest: xxx
  mode: 0755
```

will4j:
> This approach isn’t ideal because the mode parameter for copy or template no longer takes effect.like
> 
> ```yaml
> copy:
>   src: xxx
>   dest: xxx
>   mode: 0755
> ```

this only affect directory permission when call `os.MkdirAll`, the file copied by still use mode from `copyArgs`. Or Did you mean when `src` is a directory, the `mode` arg will be ignored? that's indeed a question.

will4j:
> > This approach isn’t ideal because the mode parameter for copy or template no longer takes effect.like
> > ```yaml
> > copy:
> >   src: xxx
> >   dest: xxx
> >   mode: 0755
> > ```
> 
> this only affect directory permission when call `os.MkdirAll`, the file copied by still use mode from `copyArgs`. Or Did you mean when `src` is a directory, the `mode` arg will be ignored? that's indeed a question.

And wtf! I test copy with `src` as directory and it event not work at all, I'll make another commit to fix it. Just can't beleive how this could happen, it is the basic use case of **the** `Copy` Module.

zuoxuesong-worker:
I appreciate the suggestion, but fixing directory permissions to 777 would create security issues while reducing flexibility. The Unix permission model intentionally separates file and directory permissions because they serve different purposes.

Key concerns with 777 directories:

1. Security vulnerability: Allows any user to delete or replace files regardless of file permissions
2. Violates principle of least privilege
3. Inappropriate for most real-world scenarios (config files, user data, web content)

Better approach:

- Keep the current flexible design where callers specify both permissions
- Consider adding secure defaults (e.g., 0755 for directories)
- Allow explicit 777 only when truly needed (temporary world-writable spaces)

This maintains security while supporting all legitimate use cases. For truly exceptional cases requiring 777, callers can explicitly pass 0777 as the directory mode parameter.

will4j:
I just want to fix the permission problem when copy files using `local_connector`, and respect your concern about security stuff, so I create an alias const of `0755` and use it instead. btw, `os.ModePerm` is widely used in kubkey, which may be better managed in the future.

From my aspect of view, `mode` arg in copy module should only affect the dest file permission when src is a file, and only affect dest dir permission when src is a directory. and both

```yaml 
# this is example in kubekey documents
- name: copy dir
  copy:
    src: /tmp
    dest: /tmp
```

and

```yaml
# this is example that only work in current main branch, the example above in docs did not work expectly
- name: copy dir
  copy:
    src: /tmp
    dest: /tmp/
```
should work, so I fixed the rel path problem, add a func to ensure dest dir permission after all files are copied into it.

let me know if the new impl meet the need.

redscholar:
> I just want to fix the permission problem when copy files using `local_connector`, and respect your concern about security stuff, so I create an alias const of `0755` and use it instead. btw, `os.ModePerm` is widely used in kubkey, which may be better managed in the future.
> 
> From my aspect of view, `mode` arg in copy module should only affect the dest file permission when src is a file, and only affect dest dir permission when src is a directory. and both
> 
> ```yaml
> # this is example in kubekey documents
> - name: copy dir
>   copy:
>     src: /tmp
>     dest: /tmp
> ```
> 
> and
> 
> ```yaml
> # this is example that only work in current main branch, the example above in docs did not work expectly
> - name: copy dir
>   copy:
>     src: /tmp
>     dest: /tmp/
> ```
> 
> let me know if the new impl meet the need.

I agree. `mode` only affect for `dest`

PR #2882, opened 2025-12-03
fix: set hostname failed because of unmatch variable use

will4j:
### What type of PR is this?

/kind bug

1. `.set_hostname` variable is used in roles `when` expression, but actual variable is `.native.set_hostname`.
2. mark task as **skip** when all loop is skipped to indicates some unexpected conditions.

Fixes #

```
```

```release-note
none
```

```docs

```


redscholar:
I have a new idea for the loop. The result should support scenarios both with and without an item. Once I’m done, I’ll update your PR.

will4j:
> I have a new idea for the loop. The result should support scenarios both with and without an item. Once I’m done, I’ll update your PR.

This should be fully tested, execute tasks will fail now using this new patch.

redscholar:
> > I have a new idea for the loop. The result should support scenarios both with and without an item. Once I’m done, I’ll update your PR.
> 
> This should be fully tested, execute tasks will fail now using this new patch.

I’m working on it.

mumuhhh:
> > > I have a new idea for the loop. The result should support scenarios both with and without an item. Once I’m done, I’ll update your PR.
> > 
> > 
> > This should be fully tested, execute tasks will fail now using this new patch.
> 
> I’m working on it.

由于结构性改变正常运行需要删除历史运行runtime状态目录，循环loop的registry行为了发生了变化,多item的任务受到影响，目前看hook目录下那俩*_install.yaml将会受到影响

PR #2907, opened 2025-12-24
Kubeadm conf v1beta4

will4j:
### What type of PR is this?

/kind feature

According to [this k8s doc](https://kubernetes.io/docs/reference/config-api/kubeadm-config.v1beta4/), the main change of kubeadm Configuration (v1beta4) is _Replace the existing string/string extra argument maps with structured extra arguments that support duplicates. The change applies to ClusterConfiguration - apiServer.extraArgs, controllerManager.extraArgs, scheduler.extraArgs, etcd.local.extraArgs. Also to nodeRegistration.kubeletExtraArgs._

so I add a template function to convert extraArgs config from map to slice, keep kubekey config schema unchanged.

and use v1beta4 when kube version greater than v1.31.0

Fixes #

```
```

```release-note
Use kubeadm Configuration (v1beta4) when kube version higher than v1.31.0
```

```docs

```

### DIFF
--- a/builtin/core/roles/kubernetes/init-kubernetes/templates/kubeadm/kubeadm-init.v1beta4
+++ b/builtin/core/roles/kubernetes/init-kubernetes/templates/kubeadm/kubeadm-init.v1beta4
@@ -73,7 +73,7 @@ apiServer:
       value: /etc/kubernetes/audit/webhook.yaml
 {{- end }}
 {{- if .kubernetes.apiserver.extra_args | empty | not }}
-{{ kubeExtraArgs .kubernetes.apiserver.extra_args | toYaml | indent 4 }}
+{{ mapToNamedStringArgs .kubernetes.apiserver.extra_args | toYaml | indent 4 }}
 {{- end }}
   certSANs:
     - localhost
@@ -143,7 +143,7 @@ controllerManager:
   {{- end }}
 {{- end }}
 {{- if .kubernetes.controller_manager.extra_args }}
-{{ kubeExtraArgs .kubernetes.controller_manager.extra_args | toYaml | indent 4 }}
+{{ mapToNamedStringArgs .kubernetes.controller_manager.extra_args | toYaml | indent 4 }}
 {{- end }}
   extraVolumes:
     - name: host-time
@@ -172,7 +172,7 @@ scheduler:
   {{- end }}
 {{- end }}
 {{- if .kubernetes.scheduler.extra_args }}
-{{ kubeExtraArgs .kubernetes.scheduler.extra_args | toYaml | indent 4 }}
+{{ mapToNamedStringArgs .kubernetes.scheduler.extra_args | toYaml | indent 4 }}
 {{- end }}
 ---
 apiVersion: kubeadm.k8s.io/v1beta4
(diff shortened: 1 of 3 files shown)


pkg/converter/tmpl/functions.go
```diff
@@ -139,3 +142,22 @@ func getStringSlice(d map[string][]string, key string) []string {
 	}
 	return nil
 }
+
+// kubeExtraArgs make kubeadm extra args of v1/beta4
+// for string/string extra argument maps, convert to structured extra arguments
+func kubeExtraArgs(input any) []map[string]string {
```
redscholar: We could use a more meaningful name, for example: mapToNamedStringArgs.

builtin/core/roles/kubernetes/init-kubernetes/templates/kubeadm/kubeadm-init.v1beta4
```diff
@@ -0,0 +1,247 @@
...
+      value: 2
+    - name: audit-log-maxsize
+      value: 200
+    - name: audit-policy-file
+      value: /etc/kubernetes/audit/policy.yaml
+    - name: audit-webhook-config-file
+      value: /etc/kubernetes/audit/webhook.yaml
+{{- end }}
+{{- if .kubernetes.apiserver.extra_args | empty | not }}
+{{ kubeExtraArgs .kubernetes.apiserver.extra_args | toYaml | indent 4 }}
```
redscholar: Good!!! Please help update kubeadm-init.v1beta2, kubeadm-init.v1beta3, kubeadm-join.v1beta2, and kubeadm-join.v1beta3 accordingly.

redscholar: @will4j 
image
The configuration for v3 is as follows; would it be better to retain this format?
like:
```yaml
kubernetes:
  apiserver:
    extra_args:
      - name: xxx
        value: xxx
```

will4j: I was intend to support both string/string format and NamedString format before. but extra_args in kubkey config default value (in builtin roles defaults) is type map, use NamedString will cause schema un match when merge kubekey config in playbook run. so I think it be better to keep use string/string format in kubekey config file for all versions. and convert to what it needed accordingly, which means, config unchanged as

```
kubernetes:
  apiserver:
    extra_args:
      - name: value
```

and only convert to NamedString in kubeadm v1beta4 like
```
apiServer:
  extraArgs:
    - name: xxx
      value: xxx
```

So kubekey config, kubeadm-init.v1beta2, kubeadm-init.v1beta3 etc  are unchanged in this case.

will4j: I am not sure if it is a good idea to make breaking change to use NamedString format in kubekey config, like 
```
kubernetes:
  apiserver:
    extra_args:
      - name: xxx
        value: xxx
```

although this can support new features in v1beta4 of support duplicates.

PR #2921, opened 2025-12-29
feat: multi version for download

redscholar:
### What type of PR is this?

/kind feature

Fixes #

```
```

```release-note
multi version for download
```

```docs
```


zuoxuesong-worker:
it will cause all download and copy fail with error.
and it would need a lot of change to download and copy 
maybe we can continue our current download task
if user need multi version binaries ,what about this example:
```
kk export artifact -c version1.yaml # download binaries for version1
kk export artifact -c version2.yaml # download binaries for version2
kk export artifact -c version3.yaml --download_image --pack # download binaries for version3, only download img and make tgz in the last run
```

PR #2988, opened 2026-01-31
fix: modify the condition for docker-compose download 

guofeng1996:
### What type of PR is this?
/kind bug

modify the condition for docker-compose download

Fixes #

```
修改artifact download任务中根据打包文件下载docker-compose的条件，依据image_registry下是否存在dockercompose_version决定是否下载docker-compose与文件存放路径
```

```release-note

```

```docs

```

### DIFF
--- a/builtin/core/roles/copy/tasks/binary.yaml
+++ b/builtin/core/roles/copy/tasks/binary.yaml
@@ -165,12 +165,12 @@
   tags: ["image_registry"]
   loop: "{{ .download.arch | toJson }}"
   when: 
-    - .image-registry.dockercompose_version | empty | not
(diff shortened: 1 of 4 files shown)


builtin/core/roles/copy/tasks/binary.yaml
```diff
@@ -165,12 +165,12 @@
   tags: ["image_registry"]
   loop: "{{ .download.arch | toJson }}"
   when: 
-    - .cri.dockercompose_version | empty | not
+    - .image-registry.dockercompose_version | empty | not
```
redscholar: should be `image_registry`

PR #3093, opened 2026-04-28
fix debian can not use in v4 and version check and clean process

zbb88888:

Fixes #

- debian can not use in v4x
- delete cluster but not clean
- version check problems
- support make build-kk-dev in local env in development process which is useful
- `files in /etc/kubekey/scripts` in not neccessary when `kk create cluster`, which will block creation process

```
```

```release-note
Support Debian on v4, fix etcd version check/downgrade prevention, and improve cluster deletion cleanup.
```

```docs

```

### DIFF
--- a/builtin/core/roles/defaults/tasks/main.yaml
+++ b/builtin/core/roles/defaults/tasks/main.yaml
@@ -50,6 +50,7 @@
     - name: Defaults | Get etcd.service LoadState and save to variable
       command: systemctl show etcd.service -p LoadState --value
       register: etcd_install_LoadState
+    # TODO: etcd_install_ActiveState has no consumers anywhere in the codebase; it can be removed once confirmed unnecessary.
     - name: Defaults | Get etcd.service ActiveState and save to variable
       command: systemctl show etcd.service -p ActiveState --value
       register: etcd_install_ActiveState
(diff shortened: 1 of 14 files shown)


builtin/core/roles/defaults/tasks/main.yaml
```diff
@@ -58,6 +58,30 @@
       register: etcd_install_version
       register_type: yaml
 
+    - name: Defaults | Normalize installed etcd version
+      set_fact:
+        etcd_installed_version: >-
+          {{- $versionOutput := .etcd_install_version.stdout | default "" | toYaml -}}
```
redscholar: `etcd_install_version` has been set to the YAML format via `register_type` on line 59.

builtin/core/roles/defaults/tasks/main.yaml
```diff
@@ -58,6 +58,30 @@
...
+          {{- $versionOutput | regexFind "etcd Version:[[:space:]]*[0-9]+\\.[0-9]+\\.[0-9]+" | trimPrefix "etcd Version: " | trim -}}
+
+- name: Defaults | Align target etcd version with installed cluster version
+  run_once: true
+  when: .groups.etcd | default list | empty | not
+  add_hostvars:
+    hosts: all
+    vars:
+      etcd:
+        etcd_version: >-
```
redscholar: It is not recommended to override **etcd.etcd_version** here.
This field should preferably be defined by the user and passed in via **config.yaml**.
If the value is set through **config.yaml** or the `--set` parameter, it will not be overwritten.

builtin/core/roles/uninstall/cri/docker/tasks/docker.yaml
```diff
@@ -19,6 +19,11 @@
         systemctl daemon-reload
         systemctl reset-failed containerd.service
 
+    - name: Docker | Terminate residual runtime shim processes
```
redscholar, line 22: This is an excellent change. In some scenarios, stopping containerd or docker does not terminate the container processes; it is necessary to stop containerd-shim.

builtin/core/roles/defaults/tasks/main.yaml
zbb88888: Thanks for the feedback @redscholar!

  The logic here is intentional to prevent etcd downgrades, which are not supported by etcd itself.

  The code flow is:
  1. Start with the user-configured version (`.etcd.etcd_version` from config.yaml or --set)
  2. Check all etcd nodes for installed versions
  3. Only override if an installed version is **higher** than the user-configured version
  4. This prevents accidental etcd downgrades that would cause cluster failures

  If all installed versions are lower or equal to the user config, the user's value is preserved.

  This is part of the 'etcd version alignment and downgrade prevention' feature mentioned in the PR description.

builtin/core/roles/uninstall/cri/docker/tasks/docker.yaml
zbb88888: thanks!

zbb88888:
@redscholar Hi, please review again! thanks!

builtin/core/roles/defaults/tasks/main.yaml
redscholar: @zbb88888 it’s has checked in https://github.com/kubesphere/kubekey/blob/8cde7206a8284debb63d1a3f68d8c08e65c8afce/builtin/core/roles/etcd/prepare/tasks/main.yaml#L34-L42

zbb88888:
@redscholar Thanks for the feedback @redscholar! please review again!

PR #3128, opened 2026-06-10
fix: dns error when zone is cn

redscholar:
### What type of PR is this?
/kind bug
Fixes a DNS configuration error that occurs when the zone is set to `cn`.
When `.image_registry.type` is empty (which can happen in certain configurations, such as when `KKZONE=cn`), the DNS task in `builtin/core/roles/native/dns/tasks/main.yaml` unconditionally generates image registry related host entries. This leads to invalid or unexpected DNS entries being written to `/etc/hosts`.
This PR adds a conditional check `{{- if .image_registry.type | empty | not }}` to ensure that image registry DNS entries are only generated when the image registry type is explicitly configured.
Fixes #
```
- The fix applies to both the DNS configuration generation and the KubeKey image registry control plane endpoint DNS configuration blocks.
- Please verify that DNS entries are correctly skipped when image_registry.type is not set.
```
```release-note
Fix DNS configuration error when zone is set to cn by skipping image registry entries when image_registry.type is not configured.
```
```docs
```

### DIFF
--- a/builtin/core/roles/native/dns/tasks/main.yaml
+++ b/builtin/core/roles/native/dns/tasks/main.yaml
@@ -37,6 +37,7 @@
       {{- end }}
     {{- end }}
     # kubekey image_registry control_plane_endpoint BEGIN
+    {{- if .image_registry.type | empty | not }}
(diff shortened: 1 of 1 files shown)


redscholar:
@gemini review

PR #3159, opened 2026-06-30
fix: flannel need cni_plugins

redscholar:
### What type of PR is this?

/kind bug

Multus delegates network setup to standard CNI plugins (like loopback, bridge, macvlan, host-local), which are provided by this package. Without these plugins in /opt/cni/bin/, multi-CNI deployments will fail. Furthermore, Kubelet itself requires the loopback CNI plugin to be present to set up the loopback interface for pods.

Fixes #

```
```

```release-note
none
```

```docs

```


redscholar:
@gemini review

PR #3162, opened 2026-07-06
  feat: improve web installer node checks and disk configuration

yongchuanzhou:
### What type of PR is this?

/kind feature
/kind cleanup

This PR introduces node storage formatting, multipath blacklist configuration, and GPU/NPU device detection for KubeKey.

- **Storage formatting**: Adds the `native/storage` built-in role that formats block devices, creates optional partitions, manages LVM volume groups/logical volumes, creates filesystems, and persists mounts to `/etc/fstab`. Supports both the inventory standard format (`storage`) and the web installer format (`kubernetes.storage_disks`).
- **Multipath blacklist**: Adds the `native/multipath` built-in role that idempotently manages `/etc/multipath.conf` blacklist rules, backs up the existing config, validates with `multipath -t`, and reloads the `multipathd` service.
- **GPU/NPU detection**: Adds PCI device scanning via `lspci` and a configurable GPU vendor classification config. The detected GPU vendor is exposed in gathered host facts and can be used by downstream roles.

Per review feedback, the storage and multipath logic is implemented as built-in playbook `command`/`template` tasks rather than dedicated Go modules, keeping the module layer OS-decoupled. The PR also restores three framework conventions:
- `task_executor.go` determines module failure only from the returned `error`.
- `setup.go` records connector failures via stdout/stderr without propagating a hard error, because `setup` has no `ignore_errors` option.
- `pkg/variable/helper.go` injects default `localhost` into the `all` group.

The GPU vendor classification sample is shipped as documentation under `docs/en/reference/gpu_vendors.yaml` and `docs/zh/reference/gpu_vendors.yaml`; the runtime lookup path is `work_dir/gather_facts/gpu_vendors.yaml`.

Fixes #

```
- The storage role separates script generation (`template`) from execution (`command`) so the generated scripts are inspectable on the target host.
- Helper shell functions are duplicated between the two storage templates; sharing them is left for a follow-up refactor.
- The combination of the `setup.go` and `task_executor.go` conventions means connector failures in `setup` are recorded in stdout/stderr but not in the task-level Error field.
- No CRD or inventory schema changes are introduced; storage and multipath settings are runtime inventory variables.
```

```release-note
Added built-in support for node storage formatting and multipath blacklist configuration via the `native/storage` and `native/multipath` roles. Added GPU/NPU device detection during fact gathering with a configurable vendor classification config.

The runtime GPU vendor classification config is read from `work_dir/gather_facts/gpu_vendors.yaml`; a sample file is provided under `docs/en/reference/gpu_vendors.yaml` and `docs/zh/reference/gpu_vendors.yaml`.
```

```docs
- docs/en/reference/storage.md
- docs/zh/reference/storage.md
- docs/en/reference/gpu_vendors.yaml
- docs/zh/reference/gpu_vendors.yaml
```


pkg/modules/setup/setup.go
```diff
@@ -39,7 +39,7 @@ func ModuleSetup(ctx context.Context, opts internal.ExecOptions) (string, string
 	// get connector
 	conn, err := opts.GetConnector(ctx)
 	if err != nil {
-		return internal.StdoutFailed, internal.StderrGetConnector, nil
+		return internal.StdoutFailed, internal.StderrGetConnector, err
```
redscholar: During host_check, if any one node is unreachable, the information for all nodes fails to be retrieved. Consequently, the UI cannot distinguish between reachable and unreachable nodes.

builtin/core/roles/native/storage/tasks/main.yaml
```diff
@@ -0,0 +1,37 @@
+---
+- name: Storage | Configure multipath blacklist
+  when: .kubernetes.multipath.enabled | default false
```
redscholar: Does this task have a dependency on .storage or .kubernetes.storage_disks? From the current implementation, it seems this task is executed only if .storage or .kubernetes.storage_disks is configured and kubernetes.multipath.enabled is true.

builtin/core/roles/precheck/meta/main.yaml
```diff
@@ -29,6 +29,10 @@ dependencies:
     tags: ["network"]
     when: .inventory_hostname | ne "localhost"
 
+  - role: precheck/storage
+    tags: ["storage"]
+    when: or (.storage | empty | not) (.kubernetes.storage_disks | empty | not)
```
redscholar: The when conditions is not necessary.

cmd/kk/app/options/web.go
```diff
@@ -23,10 +24,26 @@ type KubeKeyWebOptions struct {
 	UIPath     string
 }
 
+// HostCheckPlaybookPath returns the host-check playbook path.
+func (o *KubeKeyWebOptions) HostCheckPlaybookPath() string {
+	hostCheckPlaybook := filepath.Join(filepath.Dir(o.SchemaPath), "host_check.yaml")
```
redscholar: host_check.yaml should not be placed in the schema path directory.

config/scanner/gpu_vendors.yaml
redscholar, line 1: The file can be placed in the binary directory instead of being hardcoded in the code.

pkg/connector/helper.go
```diff
@@ -19,9 +19,150 @@ package connector
...
+	"encoding/json"
 	"strings"
+
+	"github.com/cockroachdb/errors"
+	"k8s.io/klog/v2"
+
+	_const "github.com/kubesphere/kubekey/v4/pkg/const"
 )
 
+type commandRunner interface {
```
redscholar: There’s no need to introduce a new CommandRunner interface. We can use the existing Connector interface instead.

pkg/connector/ssh_connector.go
```diff
@@ -49,10 +49,35 @@ var defaultSSHPrivateKey string
...
 func init() {
 	if currentUser, err := user.Current(); err == nil {
-		defaultSSHPrivateKey = filepath.Join(currentUser.HomeDir, ".ssh/id_rsa")
-	} else {
-		defaultSSHPrivateKey = filepath.Join(defaultSSHUser, ".ssh/id_rsa")
+		defaultSSHPrivateKey = defaultPrivateKeyPath(currentUser.HomeDir)
+	}
+}
+
+func defaultPrivateKeyPath(homeDir string) string {
```
redscholar: Currently, only the first SSH private key file is being read. A more thorough fix would be to load all private key files for passwordless authentication.

pkg/connector/ssh_connector.go
```diff
@@ -434,10 +461,12 @@ func (c *sshConnector) getHostInfo(ctx context.Context) (map[string]any, error)
...
 
-	// persistence the hostInfo
-
-	return map[string]any{
+	hostInfo := map[string]any{
 		_const.VariableOS:      osVars,
 		_const.VariableProcess: procVars,
-	}, nil
+	}
+	enrichHostInfoWithBlockDevices(ctx, hostInfo, c)
```
redscholar: It would be better to keep the same format as above, for example:

```go
// block devices
blockdevicesVars, err := blockDevicesFromLsblk(ctx, run)
if err != nil {
	klog.V(4).InfoS("skip block device gathering", "error", err)
    return err
}
hostInfo := map[string]any {
    _const.VariableOS: osVars,
    _const.VariableProcess: procVars,
    _const.VariableBlockDevices: blockdevicesVars,
}
```

pkg/const/common.go
```diff
@@ -89,6 +119,15 @@ const ( // === From GatherFact ===
...
 	VariableProcessMemory = "memInfo"
+	// VariableBlockDevices the value is block device info gathered via lsblk.
+	VariableBlockDevices = "blockdevices"
+	// VariableGPU the value is GPU info gathered via lspci.
+	VariableGPU = "gpu"
+)
+
+const (
+	// GPUVendorConfigRelPath is the relative path to GPU vendor config under workdir.
+	GPUVendorConfigRelPath = "config/scanner/gpu_vendors.yaml"
```
redscholar: It should be placed in the workdir.go file.

pkg/modules/multipath_conf/multipath_conf.go
```diff
@@ -0,0 +1,186 @@
...
+	for _, value := range values {
+		trimmed := strings.TrimSpace(value)
+		if trimmed != "" {
+			result = append(result, trimmed)
+		}
+	}
+	return result
+}
+
+func buildScript(cfg item) string {
```
redscholar: It should not be tightly coupled with bash.

redscholar:
@gemini review

yongchuanzhou:
@gemini review

redscholar:
@gemini review

PR #3168, opened 2026-07-15
feat: support global pre_install.sh and post_install.sh hooks

redscholar:
### What type of PR is this?

/kind feature
/kind api-change

This PR refines the pre/post-installation hook script mechanism to support a deterministic, three-tier execution order:

1. **Global script**: if `scripts_dir/pre_install.sh` (resp. `post_install.sh`) exists, copy it to all nodes and execute it first.
2. **Group scripts**: for each group listed in `pre_install_groups` / `post_install_groups` (in order), if the host belongs to that group and `scripts_dir/<group>/pre_install.sh` (resp. `post_install.sh`) exists, copy and execute it.
3. **Host-specific script**: execute `pre_install_<inventory_hostname>.sh` / `post_install_<inventory_hostname>.sh` last.

`add_nodes.yaml` is updated to also import `hook/post_install.yaml` so that post-install hooks run when adding nodes, consistent with `create_cluster.yaml`.

Fixes #

```
- No Go code changes; only built-in playbooks, defaults, and docs are modified.
- The previous `.pre_install` / `.post_install` config-based script list is replaced by `pre_install_groups` / `post_install_groups`. Users of the old config will need to migrate.
- Group script execution uses a shell `for` loop over the ordered group list instead of `find ... | sort`, guaranteeing the configured order even when a host belongs to multiple groups.
- The `when` path for hostname scripts in post_install.yaml is fixed to match the actual `src` path (no extra `/scripts/` segment).
```

```release-note
Support ordered global, group, and host-specific pre/post-install hook scripts. The previous `.pre_install` / `.post_install` config variables are replaced by `pre_install_groups` / `post_install_groups`; users switching to this release must update their configuration (action required).
```

```docs
- Updated docs/en/reference/playbooks/create_cluster.md to describe the new global → group → hostname hook script order.
- Updated docs/zh/reference/playbooks/create_cluster.md to describe the same order.
- Updated docs/en/reference/playbooks/add_nodes.md and docs/zh/reference/playbooks/add_nodes.md to include detailed pre/post hook flows.
```


redscholar:
@gemini review

PR #3243, opened 2026-09-11
feat(native): make OS package list configurable via native.packages

redscholar:
### What type of PR is this?

/kind feature

## What does this PR do

Make the OS packages installed during node initialization configurable through `native.packages`, and fix the package names used by the RHEL branch.

### Background / Motivation

The package list was hardcoded in `install_package.yaml` — `socat conntrack ipset ebtables chrony ipvsadm`, duplicated in the Debian and the RHEL task — so users could neither inspect it nor change it.

Two further problems came with it:

1. The RHEL task reused Debian package names. `conntrack` and `nfs-kernel-server` do not exist in RHEL repositories; the correct names are `conntrack-tools` and `nfs-utils`. KubeKey's own offline ISO list (`hack/gen-repository-iso/packages.yaml`, consumed as `.common[] + .rpms[]` / `.common[] + .debs[]`) already used the correct RPM names, so the runtime list had been out of sync with the ISO contents it ships.
2. The "already installed" check was a `case` statement with 8 branches, each using a different probe (`command -v`, `ipset list`, `ebtables -L`, `systemctl show -p LoadState`, `nft --version`), duplicated in both tasks. Any package outside that `case` — including every user-defined package — was silently dropped from the install list. The `chrony` branch probed `chronyd.service`, a unit name that does not exist on Debian (it is `chrony.service`), so `chrony` was reinstalled on every run.

### Implementation

1. The list is exposed as `native.packages.debs` and `native.packages.rpms`, declared per package manager with real distribution package names.
2. Both `PKGS` lines render from the corresponding parameter, so each distribution installs names that actually exist there.
3. The probe is reduced to a single query against the native package database (`dpkg-query -W -f='${Status}' | grep "install ok installed"` on Debian, `rpm -q --quiet` on RHEL). Once the configured names are real, the package database is authoritative and no binary-name fallback is required — which also means arbitrary user packages now work.
4. Conditional packages: `nftables` is appended when `kubernetes.kube_proxy.mode` is `nftables`; when the inventory defines a non-empty `nfs` group, the NFS server package is installed on that group's nodes (`nfs-kernel-server` / `nfs-utils`) and the NFS client package on all other nodes (`nfs-common` / `nfs-utils`), so workloads on those nodes can mount NFS volumes. On RPM, `nfs-utils` ships both roles, so no per-node split is needed.

### Key Changes

| File | What changed |
|---|---|
| `builtin/core/roles/defaults/defaults/main/02-native.yaml` | Added `native.packages.debs` / `native.packages.rpms` holding the previous defaults, with real package names per distribution |
| `builtin/core/roles/native/repository/tasks/install_package.yaml` | Both `PKGS` lines render from the new parameters; the 8-branch `case` probe is replaced by a single package-manager query; conditional `nftables` / NFS handling updated |
| `docs/en/reference/config.md`, `docs/zh/reference/config.md` | Documented the two parameters, their defaults and the conditional packages |
| `docs/en/installation/README.md`, `docs/zh/installation/README.md` | Split the OS dependency list per distribution |

## Impact

- **Affected modules**: `builtin/core` (native role, node initialization)
- **API changes**: N/A
- **Config changes**: New `native.packages.debs` / `native.packages.rpms`; defaults equal the previously hardcoded list, with `conntrack-tools` instead of `conntrack` on the RPM side
- **Dependency changes**: N/A

## Breaking Changes

None. Debian behaviour is unchanged. On RHEL, the packages passed to `yum` are now names that exist (`conntrack-tools`, `nfs-utils` instead of `conntrack`, `nfs-kernel-server`).

Fixes #

### Verification performed

- [x] Local build passes (`make kk`)
- [ ] Integration / E2E tests pass
- [x] Manual verification

### Steps to verify

1. `make kk`, then run a config that triggers the native role and inspect the rendered `PKGS` line.
2. Render both `PKGS` templates through `pkg/converter/tmpl` for four cases: no `nfs` group, an `nfs` node, a non-`nfs` node, and `kube_proxy.mode=nftables`.
3. Expected: the base list only, plus `nfs-kernel-server` on an `nfs` node, plus `nfs-common` on the others, plus `nftables` when the proxy mode is `nftables`.
4. On a real host, confirm already-installed packages are skipped and only missing ones reach apt/yum.

### Test coverage

Template rendering was verified for both branches across 5 scenarios (no `nfs` group; `nfs` node; k8s node; each of the latter with `nftables`), plus the case where only one of `debs` / `rpms` is configured. Both YAML files parse, and both rendered scripts pass `sh -n`. No unit test is added: the logic lives in embedded YAML playbooks and the repository has no harness for role task templates.

## Rollback

Plain revert. The change is confined to a defaults file and one task file, so `git revert` restores the previous hardcoded list.

```release-note
Add `native.packages.debs` and `native.packages.rpms` to customize the OS packages installed during node initialization. On RHEL, `conntrack-tools` and `nfs-utils` are now installed instead of the previously incorrect `conntrack` and `nfs-kernel-server`. When the inventory defines an `nfs` group, the NFS client package is installed on the remaining nodes.
```

## Checklist

- [ ] Tests added or updated
- [x] Docs / CHANGELOG updated (if needed)
- [x] Local lint and build pass (`make kk`)

```docs

```

## Notes for Reviewer

- Removing the `case` probe is what makes `native.packages.*` work for arbitrary packages; the trade-off is that package-specific semantic probes are gone, so package presence is judged by the package database only.
- `builtin/capkk/roles/init/init-os/tasks/init_repository.yaml` carries the same hardcoded list (4 occurrences). It is intentionally left untouched here.
- Pre-existing gap, not addressed here: the Debian offline ISO ships `nfs-common` but not `nfs-kernel-server`, so an `nfs` node installed from that ISO would still fail to obtain the server package.

### DIFF
--- a/builtin/core/roles/defaults/defaults/main/02-native.yaml
+++ b/builtin/core/roles/defaults/defaults/main/02-native.yaml
@@ -23,3 +23,25 @@ native:
   #   [hostname of the node being installed] -> corresponding node IP
   localDNS:
     - /etc/hosts
+  # OS packages installed on every node during initialization, declared per package
+  # manager with their real distribution package names.
+  # Additional packages are appended automatically when required:
+  #   - nftables when kubernetes.kube_proxy.mode is "nftables"
+  #   - when the inventory defines a non-empty "nfs" group: nfs-kernel-server (deb)
+  #     or nfs-utils (rpm) on the nodes of that group, and nfs-common (deb) or
+  #     nfs-utils (rpm) on all other nodes
+  packages:
+    debs:
+      - socat
+      - conntrack
+      - ipset
+      - ebtables
+      - chrony
+      - ipvsadm
+    rpms:
+      - socat
+      - conntrack-tools
+      - ipset
+      - ebtables
+      - chrony
+      - ipvsadm
--- a/builtin/core/roles/native/repository/tasks/install_package.yaml
+++ b/builtin/core/roles/native/repository/tasks/install_package.yaml
@@ -12,39 +12,17 @@
 - name: Repository | Initialize Debian-based repository and install required system packages
   command: |
     now=$(date +"%Y%m%d%H%M%S")
-    PKGS="socat conntrack ipset ebtables chrony ipvsadm{{ if .kubernetes.kube_proxy.mode | default "iptables" | eq "nftables" }} nftables{{ end }}{{ if .groups.nfs | default list | has .inventory_hostname }} nfs-kernel-server{{ end }}"
+    # nftables is appended when kube-proxy runs in nftables mode. When the inventory
+    # defines a non-empty "nfs" group, the NFS server package is installed on the
+    # nodes of that group and the NFS client package on all other nodes.
+    PKGS="{{ .native.packages.debs | default list | join " " }}{{ if .kubernetes.kube_proxy.mode | default "iptables" | eq "nftables" }} nftables{{ end }}{{ if .groups.nfs | default list | has .inventory_hostname }} nfs-kernel-server{{ else if .groups.nfs | default list | empty | not }} nfs-common{{ end }}"
     PKGS_TO_INSTALL=""
     for pkg in $PKGS; do
-      if [ -n "$pkg" ]; then
-        case "$pkg" in
-          socat)
-            command -v socat >/dev/null 2>&1 || PKGS_TO_INSTALL="$PKGS_TO_INSTALL socat"
-            ;;
-          conntrack)
-            command -v conntrack >/dev/null 2>&1 || PKGS_TO_INSTALL="$PKGS_TO_INSTALL conntrack"
-            ;;
-          ipset)
-            ipset list >/dev/null 2>&1 || PKGS_TO_INSTALL="$PKGS_TO_INSTALL ipset"
-            ;;
-          ebtables)
-            ebtables -L >/dev/null 2>&1 || PKGS_TO_INSTALL="$PKGS_TO_INSTALL ebtables"
-            ;;
-          chrony)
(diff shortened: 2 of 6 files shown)


redscholar:
## E2E Test Results for `native.packages`

Tested `kk create cluster` end-to-end on both Ubuntu and CentOS with the `native.packages` feature.

### Test Environment
- Ubuntu 24.04.1: `i-wzodkx5c` / 172.16.0.3
- CentOS Stream 9: `i-k2ngrd6h` / 172.16.0.2
- kk binary: `v4.0.7-6-00f68b49f14ac8-dirty` (built from this PR branch, `BUILDTAGS=builtin`)
- Command: `kk create cluster -c config.yaml -i inventory.yaml --workdir <dir>`

### Results
| Platform | Result | Details |
|----------|--------|---------|
| Ubuntu 24.04.1 | ✅ Pass | Playbook total: 339, failed: 0. Node Ready, v1.37.0, containerd 2.3.4, calico all Running |
| CentOS Stream 9 | ✅ Pass | Playbook total: 339, success: 327, ignored: 12, failed: 0. Node Ready, v1.37.0, containerd 2.3.4, calico all Running |

### Feature Verification
- Default debs/rpms list installed ✅
- Config override (`vim-enhanced` on CentOS) ✅
- `nftables` conditional (`kube_proxy.mode=nftables`) ✅
- NFS conditional (skipped when no `nfs` group) ✅
- Variable precedence: config > role defaults ✅

### Notes
- CentOS Stream 9 base image lacks `tar`; installed it manually before rerun (not a feature issue).
- `ebtables` has no standalone RPM on CentOS Stream 9; the binary is provided by `iptables-nft` (`rpm -q --provides iptables-nft` includes ebtables), so `rpm -q ebtables` reports not-installed but the functionality works.

PR #3247, opened 2026-09-14
feat(network): support standard IPv4/IPv6 dual-stack clusters

redscholar:
### What type of PR is this?

/kind feature

## What does this PR do

Adds standard IPv4/IPv6 dual-stack support: the address families of a cluster are derived from `cni.pod_cidr` and applied to kubelet, kube-ovn and hybridnet, and precheck now requires the node addresses, `cni.pod_cidr` and `cni.service_cidr` to agree on those families — and rejects a CIDR value that is not one or two well-formed CIDRs before any of that is compared.

### Background / Motivation

Nothing in the playbook derived the address families of a cluster from `cni.pod_cidr`, so a cluster configured for dual-stack came up half configured:

1. **kubelet advertised one family only.** `kubelet.env` was rendered with a single-family `if / else if` chain over `internal_ipv4` / `internal_ipv6`, so on a dual-stack cluster the IPv6 address was dropped and every node published just one `InternalIP`. Pods scheduled on those nodes got no address of the other family.
2. **kube-ovn and hybridnet ignored `cni.pod_cidr`.** Their roles rendered only the registry, so both charts kept their own pod CIDR defaults (kube-ovn `10.16.0.0/16`, hybridnet a single IPv4 `init` Subnet), which no longer matched the `--cluster-cidr` the rest of the cluster was installed with.

Calico was checked as well and needs **no** change here: the tigera-operator already derives its IP pools from the cluster, see the note under *Notes for Reviewer*.

### Implementation

The address families are derived from `cni.pod_cidr` (`first` / `last` entry, IPv4 first as the Kubernetes dual-stack convention requires) and consumed by each component that has to know them:

- **kubelet** — `--node-ip` is built from the node's own addresses, and a family is added only when the node owns the address **and** the cluster uses that family. kubelet publishes *every* entry of `--node-ip` as a `NodeInternalIP` (`pkg/kubelet/nodestatus/setters.go`), while kube-proxy and the controller manager route only the families of `--cluster-cidr`, so a family known to just one side is either dropped or advertised without a covering route. Guarding on both sides keeps the rendered unit correct even if precheck is skipped (e.g. add-node).
- **kube-ovn** — `networking.NET_STACK` plus per-stack `POD_CIDR` / `SVC_CIDR`; `POD_GATEWAY` is left empty so kube-ovn derives the gateway from the CIDR.
- **hybridnet** — `defaultIPFamily` plus an `init` Subnet aligned to the first family. hybridnet allocates a dual-stack pod one IPv4 **and** one IPv6 address from a paired IPv4/IPv6 Subnet of the same Network, and the chart renders exactly one `init` Subnet, so the role additionally creates the IPv6 counterpart after the chart is installed (retried, because the `networking.alibaba.com` webhook is `failurePolicy: Fail`).
- **cilium / flannel / spiderpool / cri-dockerd** — the same families drive `ipv4.enabled` / `ipv6.enabled`, `podCidr` / `podCidrv6`, `ipam.enableIPv4` / `enableIPv6` and `--ipv6-dual-stack`. Each of these templates now splits `cni.pod_cidr` itself, next to where the value is used and only for the family it serves, so the address families have a single source of truth instead of a second copy in `cni`.

precheck gains a per-node assertion that rejects any disagreement between the node addresses and the cluster CIDRs in **both** directions, plus a `run_once` assertion that `cni.pod_cidr` and `cni.service_cidr` list the same families in the same order.

Both CIDRs are also validated *before* any of those checks ask for their address family: each one has to be one or two valid CIDRs. `ipFamily` returns an error for malformed input, and an error in a `that` expression surfaces as `failed to parse argument of that` — the message of the check that caught it is never shown — so the shape check uses a new `isCIDR` template function instead, a sibling of the existing `isIP` that wraps `net.ParseCIDR` and never fails, letting the task report the value it rejected.

### Key Changes

| File | What changed |
|---|---|
| `builtin/core/roles/kubernetes/pre-kubernetes/templates/kubelet/kubelet.env` | `--node-ip` is built from the node addresses, each family gated on the cluster also using it; omits `--node-ip` entirely when nothing matches |
| `builtin/core/roles/cni/kubeovn/templates/values.yaml` | `networking.NET_STACK` and per-stack `POD_CIDR` / `POD_GATEWAY` / `SVC_CIDR` |
| `builtin/core/roles/cni/hybridnet/templates/values.yaml` | `defaultIPFamily` and the `init` Subnet aligned to the cluster families |
| `builtin/core/roles/cni/hybridnet/templates/init-ipv6-subnet.yaml` | **New** — IPv6 counterpart of the chart's `init` Overlay Subnet |
| `builtin/core/roles/cni/hybridnet/tasks/main.yaml` | Creates the IPv6 Subnet after the chart install, retrying until the webhook is reachable |
| `builtin/core/roles/cni/cilium/templates/values.yaml`, `builtin/core/roles/cni/flannel/templates/values.yaml`, `builtin/core/roles/cni/multi-cni/spiderpool/templates/values.yaml`, `builtin/core/roles/cri/cridockerd/templates/cri-dockerd.service` | derive the IPv4/IPv6 switches and the per-family CIDRs from `cni.pod_cidr` instead of reading the removed `cni.ipv*_support` / `cni.ipv*_pod_cidr` helpers |
| `builtin/core/roles/precheck/network/tasks/main.yaml` | Node addresses must match the pod CIDR families in both directions; pod and service CIDRs must use the same families; each CIDR must be one or two valid CIDRs (non-empty, at most two entries, every entry a CIDR); the IPv6 address-space assertion now reads `cni.ipv6_mask_size` where it read `cni.ipv4_mask_size`; the two pre-existing CIDR assertions are corrected |
| `pkg/converter/tmpl/functions.go` | **New** `isCIDR` template function — `net.ParseCIDR`, returns a bool and never fails, so a check can report a malformed value instead of aborting the render |
| `pkg/converter/tmpl/template_test.go` | `isCIDR` cases: IPv4 / IPv6 CIDR, the second entry of a dual-stack value, an empty entry, a bare IP, garbage |
| `test/builtin/core/dual_stack_precheck_test.go` | **New** — replays every `when` / `that` of the precheck network task file (read from the playbook, not copied) against ten inventories |
| `builtin/core/roles/defaults/defaults/main/04-cni.yaml` | The per-family helpers are removed; comments document the agreement requirement |
| `builtin/core/roles/defaults/vars/v1.24.yaml` | Drops the leftover `ipv6_support` / `type` pair under `cri`, which no template reads |
| `docs/en/reference/config.md`, `docs/zh/reference/config.md` | Dual-Stack section: the agreement rule and the per-CNI wiring table |

## Impact

- **Affected modules**: `builtin/core` (kubelet, kubeovn, hybridnet, cilium, flannel, spiderpool, cri-dockerd, precheck)
- **API changes**: N/A
- **Config changes**: no new keys. `cni.pod_cidr` / `cni.service_cidr` now accept and honour a two-family value end to end, and their documented contract now includes matching the node addresses. The internal per-family helpers that `04-cni.yaml` used to publish (`cni.ipv4_support`, `cni.ipv6_support`, `cni.ipv4_pod_cidr`, `cni.ipv6_pod_cidr`) are gone — they were never part of the documented interface, every consumer derives the value from `cni.pod_cidr` itself, and setting them now has no effect.
- **Dependency changes**: N/A

## Breaking Changes

1. **kube-ovn and hybridnet pod CIDR.** Both roles now render the pod CIDR from `cni.pod_cidr`. A kube-ovn cluster therefore gets `cni.pod_cidr` (default `10.233.64.0/18`) instead of the chart default `10.16.0.0/16`, and the hybridnet `init` Subnet moves from `100.64.0.0/16` to `cni.pod_cidr`. This is the intended behaviour — the charts previously disagreed with the `--cluster-cidr` the cluster was installed with — but it changes the pool a newly created cluster gets. On a dual-stack cluster hybridnet also gets the extra IPv6 Subnet described above.
2. **precheck is stricter.** A node that defines `internal_ipv6` while `cni.pod_cidr` is IPv4-only used to be tolerated (the address was silently unused for cluster networking); it now fails precheck, with a message naming both remedies. Conversely a node missing a family the cluster uses now fails instead of producing a half-addressed cluster.
3. **A malformed CIDR now fails with a readable message.** `cni.pod_cidr` / `cni.service_cidr` values such as `foo`, `10.233.64.1` (no prefix), `10.233.64.0/18,` (trailing comma) or `10.233.64.0/18, fd00::/64` (space after the comma) used to pass the cluster-level checks; the first three only blew up later as a template error from a node-level check, and the last one produced a CIDR with a leading space downstream. All of them now fail the new check, with the offending value in the message.
4. **The IPv6 address-space check is now effective.** It compared `max_pods` against `2^(128 - cni.ipv4_mask_size)` — the IPv4 mask — so with the default `/24` it was `2^104` and passed unconditionally. It now uses `cni.ipv6_mask_size`, so a small IPv6 per-node mask (for example `/124` with `max_pods: 110`) is rejected as it should be.

None of these apply to an installed cluster: they are all create-time configuration.

None.

### Verification performed

- [x] Unit tests pass (`go test -tags builtin ./test/builtin/core/...` — including the precheck replay below, `go test ./pkg/converter/...`, `go test ./pkg/variable/...`)
- [ ] Integration / E2E tests pass (not run — see below)
- [x] Manual verification

### Steps to verify

1. `go build ./...`, `go test ./pkg/converter/... ./pkg/variable/...` and `go test -tags builtin ./test/builtin/core/...` — pass on the rebased base.
2. Render the modified templates through KubeKey's own template engine (`pkg/converter/tmpl.Parse`) and evaluate the precheck `when` / `that` expressions through the executor's own path (`api/project/v1.ParseTmplSyntax` + `tmpl.ParseBool`) for eight scenarios: IPv4-only; dual-stack with both node addresses; dual-stack with an IPv4-only node; IPv6-only; IPv4-only cluster with a dual-stack node; IPv6-only cluster with a dual-stack node; dual-stack pod CIDR with an IPv4-only service CIDR; a custom IPv4-only CIDR.
3. The rendering context deliberately defines **no** `cni.ipv4_support` / `cni.ipv6_support` / `cni.ipv4_pod_cidr` / `cni.ipv6_pod_cidr`, so a leftover reference would show up as an unresolved `<no value>` in the rendered output; the harness asserts there is none, and that the rendered CNI / CRI values parse as YAML.
4. Expect `--node-ip` to carry exactly the families both the node and the cluster use (`172.16.0.3` / `172.16.0.3,fd85::3` / `fd85::3`), cilium / flannel / spiderpool to enable exactly the families the cluster uses and carry the matching CIDR, kube-ovn to report `NET_STACK: dual_stack` with both CIDRs, hybridnet `defaultIPFamily: DualStack` plus the IPv6 Subnet, `--ipv6-dual-stack` on cri-dockerd only when the cluster uses IPv6, and the precheck node-address verdict to be `false` for exactly the three inconsistent scenarios.
5. Evaluate the new CIDR shape assertions (same `tmpl.ParseBool` path as `assert`) for ten `cni.pod_cidr` values and compare them with the previous assertions. New: `foo`, `10.233.64.1`, `10.233.64.0/18,`, `10.233.64.0/18, fd00::/64`, `""` and a three-family value all fail — cleanly, as `false`, so the `fail_msg` is shown. Old: only `""` and the three-family value failed; for `foo` the previously reachable path was an `ipFamily` error, i.e. `failed to execute template … error calling ipFamily: foo is not ip or cidr`.
6. `go test -tags builtin ./test/builtin/core/... -run TestPrecheckNetworkAddressFamilies -v` — `test/builtin/core/dual_stack_precheck_test.go` replays the **real** `builtin/core/roles/precheck/network/tasks/main.yaml`: it reads the task file out of the embedded builtin project, flattens blocks, and evaluates every `when` and every `that` in file order through the same `kkprojectv1.ParseTmplSyntax` + `tmpl.ParseBool` path the assert module uses. Ten inventories are covered. IPv4-only with the stock defaults (`pod_cidr: 10.233.64.0/18`), IPv4-only with a custom range, IPv6-only and dual-stack each fail **no** check; the six disagreement scenarios fail on exactly the intended one (dual-stack cluster with an IPv4-only or an IPv6-only node → "Require both IPv4 and IPv6 node addresses"; IPv4-only / IPv6-only cluster with a node carrying the other family → the matching "…only" guard; dual-stack `pod_cidr` with an IPv4-only `service_cidr` → "Ensure pod CIDR and service CIDR use the same address families"; no node address at all → the pre-existing "either internal_ipv4 or internal_ipv6" check). This is the regression guard for the "at most two CIDRs" assertion: it is `le (len ...) 2`, i.e. one entry is accepted, so a single-stack cluster is unaffected.

Result: all assertions pass, no failures.

### Test coverage

Two kinds of coverage are added:

- `isCIDR` — six cases added to `TestParseCustomFunction` (`pkg/converter/tmpl/template_test.go`), run by the default `go test ./...`.
- `TestPrecheckNetworkAddressFamilies` (`test/builtin/core/dual_stack_precheck_test.go`) — the precheck replay above. Because the conditions are read from the playbook instead of copied, editing the precheck file makes this test follow it; the expectation per inventory is the exact list of failing checks. It lives under `//go:build builtin` like the other `test/builtin/core` cases, so it runs with `go test -tags builtin ./test/builtin/core/...` (`make test` alone does not pass the tag).

The rendering verification of the CNI / CRI templates (step 2-4) was done with a local harness kept outside the repository; dedicated `test/builtin/core` cases for those template renders are planned separately.

## Rollback

Plain revert. The change is create-time configuration only — no cluster state, CRD or migration is involved, and an already-running cluster is unaffected because none of the modified templates are re-applied to it.

Yes — see **Breaking Changes** above.

```release-note
KubeKey now supports standard IPv4/IPv6 dual-stack clusters. Set `cni.pod_cidr` and `cni.service_cidr` to both address families (IPv4 first) and give every node both `internal_ipv4` and `internal_ipv6`. kubelet, kube-ovn and hybridnet receive the derived dual-stack configuration; cilium, flannel, spiderpool and cri-dockerd derive it from `cni.pod_cidr` themselves, and Calico needs no change because the tigera-operator already derives its pools from the cluster. Action required: kube-ovn and hybridnet now take their pod CIDR from `cni.pod_cidr` instead of their chart defaults, precheck rejects a node whose address families do not match the cluster CIDRs, and the undocumented `cni.ipv4_support` / `cni.ipv6_support` / `cni.ipv4_pod_cidr` / `cni.ipv6_pod_cidr` helpers are no longer part of the configuration.
```

## Checklist

- [x] Tests added or updated (`isCIDR` cases in `TestParseCustomFunction`)
- [x] Docs / CHANGELOG updated (if needed)
- [x] Local lint and build pass (`make build`)

```docs
- [Kubernetes dual-stack]: https://kubernetes.io/docs/concepts/services-networking/dual-stack/
- [KubeKey reference configuration]: docs/en/reference/config.md (section "Dual-Stack (IPv4/IPv6) Configuration")
```

## Notes for Reviewer

- **Review the precheck strictness first.** "The node addresses and the cluster CIDRs must agree" is enforced in both directions, so a node that defines `internal_ipv6` while `cni.pod_cidr` is IPv4-only now fails instead of having that address silently ignored for cluster networking. That is the requested behaviour, but it is the one item here that can reject a configuration that used to pass, so it is worth an explicit ack. `internal_ipv6` is also consumed by the `/etc/hosts` generation (`native/init`, `native/dns`), so if an IPv6 address set purely for name resolution is considered legitimate, this assertion is the one to relax.
- **Calico is deliberately untouched.** The tigera-operator already derives its IP pools from the cluster: it reads the `podSubnet` (and `serviceSubnet`) of the `kubeadm-config` ConfigMap and creates one IP pool per address family, IPv4 and IPv6 alike, and it defaults the IPv6 node address detection itself. KubeKey renders that ConfigMap from `cni.pod_cidr`, so a dual-stack value already reaches Calico without any change here. This holds for every operator version KubeKey pins (`pkg/controller/installation/pod_cidr_detection.go` in v1.29.6, `pkg/controller/ippool/{kubeadm,defaults}.go` from v1.34.0). An earlier revision of this PR also emitted an explicit `ipPools` block; that was dropped again because it only duplicated the operator's own defaulting while changing three of its defaults (`blockSize`, IPv6 `natOutgoing`, the IPv6 autodetection method) as a side effect.
- **`isCIDR` is strict about whitespace.** `10.233.64.0/18, fd00::/64` is rejected rather than trimmed: nothing downstream trims either, so accepting it here would let a value through that renders as a CIDR with a leading space in `kubelet.env` and the CNI values. If tolerance is preferred, the trim has to be applied in every consumer, not in the check.
- **Also corrected, pre-existing:** the two upstream CIDR assertions at `precheck/network/tasks/main.yaml` used the *pipelined* form of `ge`, which in Go templates evaluates as `N >= value` rather than `value >= N` (`len=1` satisfies `| len | ge 2`). They were therefore asserting "at most two families" while claiming "must include both". They are now the equivalent, unambiguous `le (len ...) 2` form, keeping the existing behaviour and fixing the misleading intent. The same trap was audited across `builtin/`; every other `ge` / `le` / `gt` / `lt` pipeline found is correct.
- **capkk (Cluster API) is deliberately out of scope.** The kubekey config key `kubernetes.networking.pod_cidr` that the capkk precheck reads is not set by the released cluster-template; a Cluster API cluster declares its ranges in `Cluster.spec.clusterNetwork` (`cluster_network.pods.cidrBlocks`), which is what kubeadm, the kubelet env template and the capkk CNI values already use. Deciding the families from the former rejects a cluster created as dual-stack the standard CAPI way, so the capkk files were reverted here and will be handled separately.
- **hybridnet** is the CNI here that could not be verified statically: `defaultIPFamily: DualStack` without the paired Subnet leaves every default pod without an address, so the extra Subnet creation is the part that needs an on-cluster run. E2E on a real dual-stack cluster has not been performed yet.


redscholar:
### E2E verification: dual-stack scenarios C00–C07 (2026-09-19)

Full E2E matrix for this PR on a QingCloud **pek3** all-in-one node (Ubuntu 24.04, dual-stack: v4 `172.16.0.2` + v6 `2402:e7c0:0:6700:ffff:ffff:fffe:fffe`), kk built from this PR tip `01a370a4e` (`v4.0.7-15-01a370a4e`, `-tags builtin`), K8s **v1.37.0**, `zone: cn`.

| Scenario | CNI | Result |
|---|---|---|
| C00 | — | ✅ environment prep (node, kk build, dual-stack inventory) |
| C01 | — | ✅ precheck network negative cases **8/8** (a–h) |
| C02 | single-stack IPv4 + calico | ✅ **8/8** |
| C03 | single-stack IPv6 + calico | ✅ **8/8** (F-01 + certSANs fixed in this PR, 2026-09-20 rerun) |
| C04 | dual-stack + calico | ✅ **11/11** (2026-09-21 rerun; assertion 11 = doc-expectation error O-01, not a PR bug) |
| C05 | dual-stack + cilium | ✅ **9/9** (2026-09-19 rerun, pure `zone: cn`) |
| C06 | dual-stack + kubeovn | ✅ **7/7** (F-02/F-05/F-07 workarounds) |
| C07 | dual-stack + hybridnet | 🗑️ **removed with this PR** (F-09; upstream unmaintained since 2024-09) |

---

#### C01 — precheck network negative cases: 8/8 passed

Each negative case runs `kk precheck network` as an independent process (the executor aborts on the first failed task, so one config per case). All 8 rejected the invalid config with the expected error:

| Case | Config | Expected rejection |
|---|---|---|
| a | `cni.pod_cidr: foo` | invalid CIDR shape |
| b | `cni.pod_cidr: 10.233.64.1` (IP without mask) | invalid CIDR shape |
| c | `cni.pod_cidr: 10.233.64.0/18,` (trailing comma) | invalid CIDR shape |
| d | `cni.pod_cidr: "10.233.64.0/18, fd00:10:244::/56"` (space after comma) | strict no-trim parsing |
| e | three-family `cni.pod_cidr` | family count > 2 |
| f | pod dual-stack vs service single-stack v4 | pod/service family mismatch |
| g | dual-stack cluster, node missing `internal_ipv6` | node family ↔ cluster family mismatch |
| h | single-stack (v4-only) cluster, node has extra IPv6 | node family ↔ cluster family mismatch |

This validates the new `isCIDR` shape check and the three-way mutual-exclusion guards in `precheck/network/tasks/main.yaml`.

#### C02 — single-stack IPv4 regression (calico): 8/8 passed

Proves the PR did not break single-stack. Key assertions: kubelet dropin has **exactly one** `--node-ip` (the `{{- if }} / {{- if }}` rewrite), node addresses v4-only, `podSubnet: 10.233.64.0/18` / `serviceSubnet: 10.233.0.0/18` (no commas), exactly **1** calico IPPool (`10.233.64.0/18`, no `192.168.0.0/16` fallback), `NOTREADY=0`, coredns single IPv4 ClusterIP.

#### C03 — single-stack IPv6 (calico): 8/8 passed (2026-09-20 rerun)

Originally blocked by **F-01** (pre-existing main bug: the kubeadm template rendered `bind-address: ::` unquoted → YAML parse error → `kubeadm init` failed). **Fixed in this PR** with two commits:

- `b7040ae5` — quote the IPv6 unspecified address in both kubeadm templates (v1beta3: apiServer/controllerManager/scheduler; v1beta4: same three): `value: ::` / `bind-address: ::` → `"::"`.
- `dd7b17c1` — quote the first certSAN entry: on a pure-IPv6 cluster it renders the service CIDR network address (`fd00:10:96::`), which ends with `::`; unquoted, the trailing `:` is parsed as a mapping separator and the entry becomes a map → kubeadm fails with `cannot unmarshal object into Go struct field APIServer.apiServer.certSANs of type string`.

Rerun with kk `v4.0.7-17-dd7b17c1` (pure `zone: cn`, inventory with only `internal_ipv6`):

| # | Assertion | Result |
|---|---|---|
| 1 | precheck EXIT | ✅ `0` |
| 2 | create EXIT | ✅ `0` (337/334/3/0) |
| 3 | kubelet dropin | ✅ `--node-ip=2402:e7c0:0:6700:ffff:ffff:fffe:fffe` (v6 only, no v4) |
| 4 | node addresses | ✅ only `InternalIP=2402:e7c0:0:6700:ffff:ffff:fffe:fffe` |
| 5 | kubeadm-config | ✅ `podSubnet: fd00:10:244::/56`, `serviceSubnet: fd00:10:96::/112` |
| 6 | coredns clusterIPs | ✅ single IPv6 `fd00:10:96::2` |
| 7 | ROLLOUT_EXIT | ✅ `0` |
| 8 | v6-probe podIPs | ✅ exactly 1 IPv6 per pod, both in `fd00:10:244::/56` |

Environment note (not this PR): the QingCloud node has **no IPv6 default route** (the `inet6 dhcp` stanza is commented out; the v6 address is added manually), so on a pure-IPv6 cluster the node's `connect()` to a ClusterIP fails route lookup with `network is unreachable` before kube-proxy's nftables OUTPUT rules can DNAT it → tigera-operator can't reach the apiserver → calico never installs. Worked around with `ip -6 route add fd00:10:96::/112 dev eth0` (temporary). Dual-stack is unaffected (v4 has a default route). Worth a precheck warning for v6-only clusters.

#### C04 — dual-stack + calico: 10/11 passed

| # | Assertion | Result |
|---|---|---|
| 1 | precheck EXIT | ✅ `0` (29/27/2/0) |
| 2 | create EXIT | ✅ `0` (345/334/11/0) |
| 3 | kubelet dropin (**PR core**) | ✅ `--node-ip=172.16.0.2,2402:e7c0:0:6700:ffff:ffff:fffe:fffe` (comma-separated, v4 first) |
| 4 | node addresses | ✅ 2 `InternalIP` (v4 + v6) |
| 5 | kubeadm-config | ✅ `podSubnet: 10.233.64.0/18,fd00:10:244::/56`, `serviceSubnet: 10.233.0.0/18,fd00:10:96::/112` |
| 6 | calico ippools | ✅ 2 rows: `default-ipv4-ippool 10.233.64.0/18` + `default-ipv6-ippool fd00:10:244::/56` |
| 7 | NOTREADY | ✅ `0` |
| 8 | ROLLOUT_EXIT | ✅ `0` |
| 9 | ds-probe podIPs | ✅ 1 v4 + 1 v6 per pod |
| 10 | ds-probe Service | ✅ `PreferDualStack` / `[IPv4 IPv6]` / v4+v6 ClusterIPs |
| 11 | coredns | ✅ `SingleStack IPv4 10.233.0.3` — **O-01, doc-expectation error, not a PR bug** (2026-09-21 rerun: same result, doc fixed) |

Assertion 11: kk pins the DNS ClusterIP itself via `{{ index (.cni.service_cidr | ipInCIDR) 2 }}`; `ipInCIDR` concatenates v4 addresses first, so index 2 is always the 3rd IPv4 → coredns is single-stack IPv4 in dual-stack clusters by existing design. PR #3247 does not touch DNS templates (`git diff` on `templates/dns/` is empty). Observation (non-blocking): IPv6-only pods cannot resolve DNS in a dual-stack cluster — left for the PR author to decide.

#### C05 — dual-stack + cilium: 9/9 passed (2026-09-19 rerun, pure `zone: cn`)

Re-ran after the missing `cilium/operator-generic` tags were synced into `hub.kubesphere.com.cn`. Config: `cni.type: cilium`, pod CIDR `10.233.64.0/18,fd00:10:244::/56`, service CIDR `10.233.0.0/18,fd00:10:96::/112`, `ipv4_mask_size: 24`, `ipv6_mask_size: 64`.

| # | Assertion | Result |
|---|---|---|
| 1 | precheck exit | ✅ `0` |
| 2 | create exit | ✅ `0` (337 tasks, 335 success, 2 ignored, 0 failed) |
| 3 | `cilium-default-values.yaml` render (12/12 exact grep) | ✅ `ipv4.enabled: true`, `ipv6.enabled: true`, `clusterPoolIPv4PodCIDRList: [10.233.64.0/18]` + mask `24`, `clusterPoolIPv6PodCIDRList: [fd00:10:244::/56]` + mask `64` |
| 4 | `cilium-config` ConfigMap | ✅ `enable-ipv4: true`, `enable-ipv6: true`, cluster-pool v4/v6 CIDR + mask `24`/`64` |
| 5 | kubelet dropin | ✅ `--node-ip=172.16.0.2,2402:e7c0:0:6700:ffff:ffff:fffe:fffe` |
| 6 | NOTREADY | ✅ `1` (only the 2nd `cilium-operator` replica Pending — chart default `replicas: 2` + hostname anti-affinity, expected on a single node) |
| 7 | ds-probe rollout | ✅ `0` |
| 8 | ds-probe pod IPs | ✅ 2 IPs per pod (v4 + v6); Service `PreferDualStack` with v4+v6 ClusterIPs |
| 9 | `cilium status` | ✅ `OK` — `Cilium: Ok 1.20.1`, IPAM IPv4+IPv6, Masquerading IPv4+IPv6 Enabled, 38/38 controllers healthy, cluster health 1/1 |

#### C06 — dual-stack + kubeovn: 7/7 passed

Hit three pre-existing kk bugs (identical in main and this PR — not regressions), worked around without changing CNI semantics:

| Bug | Blocked at | Workaround |
|---|---|---|
| **F-02** whitelist typo | precheck: `require_network_plugin` says `'kube-ovn'` but `cni.type` only accepts `kubeovn` → create blocked | override via `cluster_require.require_network_plugin` in config |
| **F-05** chart filename | download lands `cni/kubeovn/kube-ovn-<v>.tgz`, role looks for `kubeovn-<v>.tgz` | background watcher `cp -n` byte-identical copy |
| **F-07** helm namespace | `helm upgrade --install --namespace kubeovn-system` missing `--create-namespace` → `namespaces "kubeovn-system" not found` | watcher pre-creates the ns once apiserver is ready |

`CREATE_EXIT=0` (345/343/2/0). Assertions: precheck `0`; create `0`; `kubeovn-default-values.yaml` six key lines all OK (`NET_STACK: dual_stack`, `POD_CIDR: 10.233.64.0/18,fd00:10:244::/56`, `SVC_CIDR: 10.233.0.0/18,fd00:10:96::/112`, no `ipv6:` section); kubelet dropin dual `--node-ip`; `NOTREADY=0` (14 Running); rollout `0`; each ds-probe pod 1 v4 + 1 v6. Extra evidence: `subnets ovn-default` `cidrBlock=10.233.64.0/18,fd00:10:244::/56` `protocol=Dual`, `join` subnet dual too, helm release `kubeovn` @ `kubeovn-system` `deployed` (chart `kube-ovn-v1.16.2`). Same O-01 coredns observation as C04.

#### C07 — hybridnet: removed with this PR

hybridnet support was **deleted as part of this PR** (2026-09-18): upstream alibaba/hybridnet has been unmaintained since 2024-09, and the PR's own `init-ipv6-subnet.yaml` used the full IPv6 pod CIDR (`/56` = 2^72) which exceeds hybridnet's hard IPAM cap of 65536 (2^16) → webhook deterministically rejects (**F-09**). The scenario, F-08/F-09 and the template were removed together; no longer executed.

---

**Environment notes (not related to this PR)**

- **O-07**: the QingCloud VPC host has `eth0.vxlan4` (vxlan id 4, dstport **8472**) which collides with cilium's default vxlan tunnel port → datapath fails to init (`address already in use`). Worked around with `tunnel-port: 8473` in `cilium-config`. Any cilium deployment on this cloud would hit the same; consider `cni.cilium.values.tunnelPort: 8473` for such environments.
- **O-08**: a stale policy-route entry in table `40000` (`10.233.64.0/18 dev eth0.vxlan4`) black-holed node→pod / pod→Service traffic; removed. Leftover from a previous kube-ovn run, not from this PR.
- **O-01**: kk pins the DNS ClusterIP from the v4 range first, so coredns is single-stack IPv4 in dual-stack clusters (existing design; PR doesn't touch DNS templates).
- Probe pods use `hub.kubesphere.com.cn/kubernetes/pause:3.10.2` + `IfNotPresent` (the sandbox image kk writes into containerd under `zone: cn`) — no external network dependency.

**Conclusion**: dual-stack works end-to-end with calico (C04), cilium (C05) and kubeovn (C06); single-stack IPv4 regression is clean (C02); precheck guards reject all 8 invalid configs (C01); and single-stack IPv6 now works too (C03, 8/8) after this PR fixed the two kubeadm-template quoting bugs (F-01 + certSANs). The PR's `values.yaml` / `kubelet.env` changes render correctly for dual-stack.

PR #3257, opened 2026-09-22
fix(cri): read the registry configuration from config_path on containerd v2

redscholar:
/kind bug

## What does this PR do

Write the registry configuration of a containerd v2 node as `hosts.toml` files under `/etc/containerd/certs.d` and point `registry.config_path` at them, because containerd v2 no longer reads `registry.configs.<host>.tls`.

### Background / Motivation

`roles/cri/containerd/templates/config.toml` describes a private registry with `registry.configs.<host>.tls` (`ca_file`, `cert_file`, `key_file`, `insecure_skip_verify`) next to the credentials. containerd v2 removed that field: `RegistryConfig` only carries `auth` now, and the daemon reports the key as unknown while still accepting it in the TOML:

```
level=warning msg="Ignoring unknown key in TOML for plugin" key="registry configs dockerhub.kubekey.local tls" plugin=io.containerd.cri.v1.images
level=warning msg="`configs` is deprecated, please use `config_path` instead"
```

The CA is therefore never handed to the resolver, kubelet falls back to the system trust store, and every pull from a private TLS registry fails:

```
failed to pull image "dockerhub.kubekey.local/kubernetes/pause:3.10.2":
rpc error: ... failed to resolve reference ...: tls: failed to verify certificate:
x509: certificate signed by unknown authority
```

On a fresh `kk create cluster` this surfaces as `kubeadm init` never reaching `wait-control-plane`. It affects every Kubernetes version whose default CRI is containerd v2 (v1.30 – v1.37, `v2.2.6` / `v2.3.3`), not only the newest one.

Extending the current layout with a `config_path` is not possible either: containerd rejects `mirrors` together with `config_path` (`containerd v1.7.13 pkg/cri/config/config.go:504`, `containerd v2.3.3 internal/cri/config/config.go:502-506`):

```
invalid cri image config: `mirrors` cannot be set when `config_path` is provided
```

### Implementation

Use the v2 native form for a v2 containerd and leave v1 untouched:

1. `registry.config_path = "/etc/containerd/certs.d:/etc/docker/certs.d"` (containerd's own default pair of directories, so certificates placed under `/etc/docker/certs.d` keep working).
2. Write one `hosts.toml` per registry under `<certs_dir>/<host>/`, carrying the mirrors, the insecure registries and the TLS material. `hosts.toml` is the only place left that can express `ca`, `client` and `skip_verify`.
3. Keep the credentials in `registry.configs.<host>.auth`: with `config_path` set, `CRIImageService.registryHosts()` (`containerd v2.3.3 internal/cri/server/images/image_pull.go:496-510`) still resolves them through `HostOptions.Credentials` → `c.config.Registry.Configs[host].Auth`, so no credential handling is lost.

The material is generated by a rendered shell script for the same reason `harbor/tasks/*` uses one: the number of registries is dynamic and each hosts.toml is written from a mixture of optional fields. The script reads the files that the registry tasks of the role already placed in the certs directory and only references the ones that are present.

Why not simply rely on the certs.d file layout without a `hosts.toml`? `loadCertsDir()` (`remotes/docker/config/hosts.go`) treats every `*.crt` file as a CA and only `*.cert` as a client certificate, so the `client.crt` that the role copies would not be used as a client key pair, and that layout has no way to express `insecure_skip_verify`.

### Key Changes

| File | What changed |
|---|---|
| `builtin/core/roles/cri/containerd/templates/config.toml` | derive `$containerdV2` from `cri.containerd_version` / the installed containerd and emit `config_path` plus `configs.<host>.auth` for v2, while v1 keeps the mirrors/configs layout |
| `builtin/core/roles/cri/containerd/templates/registry-hosts.sh` | new template: writes `hosts.toml` for `cri.registry.mirrors`, `cri.registry.insecure_registries`, `image_registry` and `cri.registry.auths` |
| `builtin/core/roles/cri/containerd/tasks/main.yaml` | render and apply the script on a v2 node, before containerd is restarted, so the config file and the hosts.toml files are in place when it starts |

## Impact

- **Affected modules**: `cri` (containerd only)
- **API changes**: N/A
- **Config changes**: N/A. For a v2 node the *generated* `/etc/containerd/config.toml` now contains `registry.config_path` instead of `registry.mirrors` and `registry.configs.<host>.tls`; no user facing item is added, changed or removed.
- **Dependency changes**: N/A

## Breaking Changes

None. containerd v1 output is byte for byte unchanged, and for v2 the previous output was already non functional for a private registry.

Fixes #

### Verification performed

- [ ] Unit tests pass (`<command>`)
- [ ] Integration / E2E tests pass (`<command>`)
- [x] Template rendering verified locally
- [x] Manual verification

### Template rendering

The new template was rendered and the generated script executed against four variable sets
(`go run` driving `pkg/converter/tmpl.Parse`, the same entry point the role uses), then the
resulting `hosts.toml` files were inspected:

| Scenario | `image_registry.auth.registry` | Result |
|---|---|---|
| mirrors + insecure registry + `cri.registry.auths` | `dockerhub.kubekey.local/library` | 4 `hosts.toml` files, each with the expected `server` / `ca` / `client` / `skip_verify` |
| default offline cluster (no mirrors, no auths, no optional auth keys) | `dockerhub.kubekey.local/library` | `server = "https://dockerhub.kubekey.local"`, `skip_verify = true` |
| registry set to the built-in mirror | `hub.kubesphere.com.cn/kubesphere` | no `hosts.toml`, the block is skipped as before |
| plain HTTP registry | `192.168.1.10:5000` | `server = "http://192.168.1.10:5000"`, `skip_verify = true` |

This is also how the optional `skip_tls_verify` is exercised: `hasKey ... | ternary (.x | toBool) true`
would evaluate `toBool` eagerly and fail the whole render when the key is absent, so the default
is taken before the override, exactly like `config.toml` does it.

### Steps to verify

On a node with containerd v2.3.3 (Kubernetes v1.37.0) and a Harbor registry reachable as `dockerhub.kubekey.local`:

1. With the current template, `crictl pull dockerhub.kubekey.local/kubernetes/pause:3.10.2` fails with `x509: certificate signed by unknown authority`, and the containerd journal shows `Ignoring unknown key ... key="registry configs dockerhub.kubekey.local tls"`.
2. The same CA used explicitly, `ctr -n k8s.io images pull --hosts-dir /etc/containerd/certs.d ...`, gets past the TLS handshake (`401` instead of a certificate error), which confirms the material and the trust configuration are valid and only the config.toml wiring is missing.
3. With this change the generated `/etc/containerd/config.toml` contains `config_path = "/etc/containerd/certs.d:/etc/docker/certs.d"` and `/etc/containerd/certs.d/dockerhub.kubekey.local/hosts.toml` exists; after `systemctl restart containerd` the pull in step 1 no longer reports a certificate error.
4. `kk create cluster` on Kubernetes v1.37.0 (containerd v2.3.3) with an HA Harbor registry: the end-to-end result will be posted as a comment on this PR.

### Test coverage

The change is a role template and a task; kubekey has no unit test harness for role templates. Verification is done on a real cluster as described above.

## Rollback

Plain revert. `containerd` regenerates `config.toml` from the template on the next run; the generated `hosts.toml` files are inert once nothing points at `config_path` and can be left in place.

```release-note
Fix image pulls from a private registry on nodes with containerd v2: the registry configuration is written as hosts.toml under /etc/containerd/certs.d and registry.config_path points at it, because containerd v2 no longer reads registry.configs.<host>.tls and rejects registry.mirrors together with config_path.
```

## Checklist

- [ ] Tests added or updated
- [ ] Docs / CHANGELOG updated (if needed)
- [x] Local build passes (`make kk`; this repository has no `make build` target)

## Notes for Reviewer

- Scope: `roles/cri/docker/templates/config.toml` has the same registry block, but with `cri.container_manager: docker` the images are pulled by dockerd, which reads `/etc/docker/certs.d` and `daemon.json`, so that block is not on the image pull path. It is intentionally left alone, and can be changed in a follow-up if you prefer both files to stay in sync.
- The v2 detection combines `cri.containerd_version` (the version this role installs) with the `--version` output of the containerd that is present on the node (`containerd/v2`), so a pre-installed v2 is detected as well.
- `hosts.toml` is also the place for `cri.registry.mirrors` now, because a v2 resolver ignores `mirrors` whenever `config_path` is set. The generated file lists the mirrors first and `registry-1.docker.io` as the fallback, which matches the previous mirror semantics.
- The `configs.<host>.auth` credentials are kept on purpose (see Implementation), so no `imagePullSecrets` migration is needed for existing setups.


redscholar:
## Cluster verification of the last three commits

`kk` was rebuilt from HEAD and checked in place on both nodes (`md5sum` on the build host and on each node, `kk version` → `v4.0.7-22-5b42c65bb4d535`, clean tree).

The earlier rounds ran against `82f08c39`. The three commits that landed after it change behaviour none of those rounds exercised, so seven cases were added for them:

| Commit | Behaviour under test |
|---|---|
| `206da14f`, `ffb3e11d` | the three `Copy image registry {CA certificate,client certificate,client key}` tasks now also require `image_registry.auth.registry` to be non-empty and not the `hub.kubesphere.com.cn` mirror |
| `206da14f`, `ffb3e11d` | the `hosts.toml` task judges v2 as `cri.containerd_version \| default "" \| hasPrefix "v2."` |
| `5b42c65b` | `$containerdV2` in `templates/config.toml` collapsed onto that same single condition |

### Results

Node A runs k8s v1.37.0 with containerd v2.3.3, node B runs k8s v1.29.15 with containerd v1.7.28 (in both cases the version the role installs).

| Case | Node | Scenario | Result |
|---|---|---|---|
| binary | A + B | rebuild from HEAD, distribute, verify | PASS — md5 identical on the build host and both nodes, and different from the previous build |
| A1 | A | `auth.registry` resolves to the `hub.kubesphere.com.cn` mirror (zone default) while `auth.ca_file` points at a cert that exists on the node | PASS — all three copy tasks skipped; no cert appears under `certs.d`, neither at its root nor in `certs.d/hub.kubesphere.com.cn`; `certs.d` holds exactly the three `hosts.toml` files |
| A2 | A | `image_registry.auth.registry` explicitly empty | PASS — the same three tasks skipped, four paths ABSENT; the `hosts.toml` files are still rewritten (mtime advances), so the cri role did reach the v2 tasks |
| B1 | A | rerun against the same workdir | PASS — the three `hosts.toml` and `config.toml` are byte-identical (md5 unchanged), node Ready, `NRestarts` 0 |
| C1 | A | target containerd pinned to v1.7.28 on a node already running v2.3.3 | PASS — `config.toml` renders the v1 layout: `registry.mirrors` and `registry.configs."harbor.internal:8443".tls` present, `config_path` absent. The same config without the pin renders `config_path` 1 / `registry.mirrors` 0. Node restored to v2.3.3 afterwards, Ready |
| D1 | B | `auth.registry` is a real registry and `auth.ca_file` exists | PASS — all three tasks run and the three certs land in `certs.d/harbor.internal:8443/`; no `hosts.toml` there, as expected on v1 |
| D2 | B | v1 layout regression | PASS — `config_path` absent, `registry.mirrors` + `registry.configs.<host>.tls` present, node Ready |

### How the skips were attributed

"Certs are absent" on its own does not prove the new guard fired — a `fileExists` that happens to be false gives the same outcome. Three checks separate the two:

- The effective values were read from the rendered precheck command: `Cert | Set cert check command` prints `check.sh <registry> --insecure --ca-file …` into the `-v 6` log. Run 1 → `check.sh hub.kubesphere.com.cn --insecure`; run 2 → `check.sh  --insecure` (two spaces, i.e. an empty host). The workdir variable snapshot is not usable for this: it does not carry the values from the config file at all.
- In the same run the path in `auth.ca_file` exists on the node, and the precheck task that would rewrite it (`Exec check shell with local certs`) was skipped, so `fileExists` evaluates true there. `fileExists` is `os.Stat`, evaluated in the `kk` process on the node.
- D1 is the positive control for A1/A2: same tasks, same guard, real registry value — the copies do run and the certs do land.

### Environment

Node A's cluster was deleted after A2 (`/etc/containerd` and `/etc/kubernetes` gone, `failed: 0`). The v1.29.15 cluster on node B is still up with the D1 certs in place, in case anyone wants to inspect it.

redscholar:
## Cluster verification of the config_policy / v1→v2 upgrade round (8559a5ce)

`kk` was rebuilt from HEAD (`8559a5ce`, clean tree; `git describe` → `v4.0.7-23-g8559a5cee`) and run on Node B. This round answers two review questions the previous seven cases (at `5b42c65b`) did not touch: (1) should the `config.toml` override / merge logic also apply to `hosts.toml`, and (2) is the v1 → v2 upgrade path tested at all.

### What this round covers (vs. the previous round at 5b42c65b)

The single new commit `8559a5ce` changes behaviour none of the earlier rounds exercised:

| Commit | Behaviour under test |
|---|---|
| `8559a5ce` | `config_policy: merge` now drops the old `registry` block (mirrors + configs) from the file being merged, so a v2 node never inherits the v1 `registry.mirrors` / `registry.configs.<host>.tls` that a v1 → v2 upgrade left behind |
| `8559a5ce` | `config_policy` now also gates the `hosts.toml` write task (same condition as `Configure containerd`), and `merge` preserves the node's existing `hosts.toml` files as whole files |
| `8559a5ce` | the v1 → v2 upgrade path itself, under `config_policy` = `""` / `merge` / `override` |

### Results

All four cases ran on Node B (k8s v1.29.15, containerd upgraded in place v1.7.28 → v2.3.3).

| Case | Node | Scenario | Result |
|---|---|---|---|
| GE1 | B | `config_policy` unset, v1 → v2 upgrade | PASS — render is the v2 layout: `grep -ac 'config_path'` = 1, `grep -ac 'registry.mirrors'` = 0; `containerd -v` → v2.3.3 |
| GE2 | B | `config_policy: merge`, v1 → v2 upgrade — reproduce the P0, then verify the fix | PASS — see the P0 note below |
| GE3 | B | `config_policy` now also governs `hosts.toml` | PASS — merge: the node's existing `hosts.toml` md5 is byte-identical before and after the rerun (`90dc2f4d65e24f4dac7330a5612dc70b` for `docker.io`, `ec144ff8247b174d95e37afd75b5a348` for `harbor.internal:8443`); override: `hosts.toml` is rewritten by the template and the node-side `skip_verify` is wiped |
| LS | B | load smoke under `config_policy` = `""` / `merge` / `override` | PASS — `crictl info` no longer reports `unknown service`, `systemctl is-active containerd` is active, node stays `Ready`, `NRestarts` = 0 in all three |

### The P0 that this round caught (GE2, pre-fix)

Under `config_policy: merge`, the upgrade merged the node's v1-written `config.toml` into the freshly rendered v2 one. The merged file carried `registry.config_path = "/etc/containerd/certs.d"` **and** the v1 `registry.configs."harbor.internal:8443".tls` (4 keys) **and** `registry.mirrors."docker.io"`. containerd v2 refuses that combination:

```
failed to load plugin ... "invalid cri image config: `mirrors` cannot be set when `config_path` is provided"
```

The cascade took `io.containerd.cri.v1.images` → `podsandbox.controller` → `sandbox-controllers` → `grpc.v1.cri` down, so `crictl info` / `crictl pull` both answered `unknown service runtime.v1.ImageService`, and the node went `NotReady`. **`kk` still reported `failed: 0`** — the containerd process itself starts, only the CRI image plugin fails to load, so this is a silent failure unless you read journald. The same node self-healed after the fix: `config_path` count = 1 / `mirrors` = 0 / `tls` = 0, `crictl pull` succeeds, node back to `Ready`.

### Environment

Node B's v1.29.15 cluster (carrying the `merge` / `override` configs) is still up in case anyone wants to inspect it. Node A's cluster was already deleted in the previous round.

redscholar:
## Cluster verification — all test cases (HEAD 8559a5ce)

`kk` was rebuilt from HEAD (`8559a5ce`, clean tree; `git describe` → `v4.0.7-23-g8559a5cee`) for the latest round and from `5b42c65b` for the earlier seven cases; both binaries were md5-verified on the nodes they ran on. Node A runs k8s v1.37.0 + containerd v2.3.3; Node B runs k8s v1.29.15 (containerd upgraded in place v1.7.28 → v2.3.3 in the latest round).

### Results

| Case | Node | Scenario | Result |
|---|---|---|---|
| binary | A + B | rebuild from HEAD, distribute, verify | PASS — md5 identical on build host and both nodes, and different from the previous build |
| A1 | A | `auth.registry` resolves to the `hub.kubesphere.com.cn` mirror while `auth.ca_file` points at a cert that exists on the node | PASS — all three copy tasks skipped; no cert under `certs.d` (root or `certs.d/hub.kubesphere.com.cn`); `certs.d` holds exactly the three `hosts.toml` files |
| A2 | A | `image_registry.auth.registry` explicitly empty | PASS — same three tasks skipped, four paths ABSENT; `hosts.toml` files still rewritten (mtime advances), so the role reached the v2 tasks |
| B1 | A | rerun against the same workdir | PASS — the three `hosts.toml` and `config.toml` are byte-identical (md5 unchanged), node Ready, `NRestarts` 0 |
| C1 | A | target containerd pinned to v1.7.28 on a node already running v2.3.3 | PASS — `config.toml` renders the v1 layout (`registry.mirrors` + `registry.configs."harbor.internal:8443".tls` present, `config_path` absent); without the pin it renders `config_path` 1 / `registry.mirrors` 0; node restored to v2.3.3, Ready |
| D1 | B | `auth.registry` is a real registry and `auth.ca_file` exists | PASS — all three tasks run, the three certs land in `certs.d/harbor.internal:8443/`; no `hosts.toml` there, as expected on v1 |
| D2 | B | v1 layout regression | PASS — `config_path` absent, `registry.mirrors` + `registry.configs.<host>.tls` present, node Ready |
| GE1 | B | `config_policy` unset, v1 → v2 upgrade | PASS — render is the v2 layout: `grep -ac 'config_path'` = 1, `grep -ac 'registry.mirrors'` = 0; `containerd -v` → v2.3.3 |
| GE2 | B | `config_policy: merge`, v1 → v2 upgrade — reproduce the P0, then verify the fix | PASS — see the P0 note below |
| GE3 | B | `config_policy` now also governs `hosts.toml` | PASS — merge: node's existing `hosts.toml` md5 byte-identical before/after rerun (`90dc2f4d65e24f4dac7330a5612dc70b` for `docker.io`, `ec144ff8247b174d95e37afd75b5a348` for `harbor.internal:8443`); override: `hosts.toml` rewritten by template, node-side `skip_verify` wiped |
| LS | B | load smoke under `config_policy` = `""` / `merge` / `override` | PASS — `crictl info` no longer reports `unknown service`, `systemctl is-active containerd` active, node Ready, `NRestarts` 0 in all three |

The "cert ABSENT" results are distinguished from a false `fileExists` by reading the rendered precheck command (`Cert | Set cert check command` prints `check.sh <registry> --insecure --ca-file …`): run 1 → `hub.kubesphere.com.cn`, run 2 → empty host (two spaces). D1 is the positive control — same tasks, real registry value, the copies do run.

### P0 caught by GE2 (pre-fix)

Under `config_policy: merge`, the upgrade merged the node's v1-written `config.toml` into the v2 one, which carried `registry.config_path` **and** v1 `registry.configs."harbor.internal:8443".tls` (4 keys) **and** `registry.mirrors."docker.io"`. containerd v2 refuses that: `invalid cri image config: mirrors cannot be set when config_path is provided`. The CRI image plugin failed to load, `crictl` answered `unknown service runtime.v1.ImageService`, and the node went `NotReady` — while `kk` still reported `failed: 0` (the containerd process starts, only the plugin fails; a silent failure unless you read journald). The same node self-healed after the fix: `config_path` = 1 / `mirrors` = 0 / `tls` = 0, `crictl pull` succeeds, node `Ready`.

### Environment

Node B's v1.29.15 cluster (carrying the `merge` / `override` configs) is still up for inspection. Node A's cluster was deleted after the earlier round.

